Issue #1, opened 2021-01-27
Suggestions based on current status.


kattni:
@FoamyGuy @kmatch98 Thank you so much for writing this! It made my project much simpler. Here are a few thoughts!

For context, I am currently working with [this file](https://gist.github.com/FoamyGuy/3d5826ede385554b54175b66c624113a).

The project I am using it in can be found [here](https://github.com/kattni/light_box/blob/main/light_box.py).

In terms of `add_sub_view`:
* The naming of the function `add_sub_view` doesn't make it clear to me what it's doing - the example did that. I do not know, however, from where the name was derived, so it could be a `displayio` thing, in which case proceed as is. Really, I am unclear on what the concept of a `view` is - this may be a concept I'm simply unfamiliar with in which case education will resolve the issue. Otherwise, consider a naming change to make things more clear.
* Along those lines, `view` could be clearer. That or we supplement with excellent documentation. :)
* `view_grid_size` should default to `(1, 1)` as that seems like a sane default. It would significantly shorten up a basic constructor.

In the `GridLayout` constructor:
* `child_padding` could be more clearly named, I had to mess with it to figure out what it did, and I'm still not sure I really know.
* `max_children` could default to the grid size `x` * `y`.
* Looking at it, I'm wondering what `child_padding` does versus what setting `x` and `y` do?

Feature requests:
* I would appreciate a way to nudge or "pad" columns or rows. I did it initially by adding spaces into all of the text elements, then went back and created a grid that included empty columns as padding. Basically I have a column on each side of the display that is only a few characters wide, and the ones in the middle are much wider. So I "padded" it with empty columns. The results were successful, but I'm wondering if there's some way to control subsections of the grid without making everything a lot more complicated. (That last bit is important!)

These are the things that occurred to me while using the version linked above in a project. I'm sure I'll come up with more as we work through it!

FoamyGuy:
On the "view" naming and concept: One thing that led to the use of "sub_view" was an attempt to avoid using parent/child relationship terms. But it turned out inconsistent, because it's still used in some places. It may be worthwhile to use them in the "add" function name to make it more clear and consistent. 

The specific use of the term "view" came from my background with Android. There is an object `View` which represents a similar concept to how I've used it here.

[...]

PR #2, opened 2021-01-28
Initial code

FoamyGuy:
Right now it's got the cookie-cutter files, the grid_layout, and everything needed to pass CI.

Still needs at least one example. I think it would also be good to maybe hold off releasing this library (or grid_layout at least) until after a few changes are merged to `display_button` and `display_text` that will make both of them work better with this library to alleviate some of the mentioned in #1


FoamyGuy:
This now has the initial code for grid_layout and a few minimal examples. I think it's ready to go at least as a starting point

adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # pylint: disable=too-many-arguments
+    def __init__(self, x, y, width, height, grid_size, child_padding, max_children=4):
+        super().__init__(x=x, y=y, max_size=max_children)
+        self.x = x
+        self.y = y
+        self._width = width
+        self._height = height
+        self.grid_size = grid_size
+        self.child_padding = child_padding
```
kattni: We could consider using `cells/cell` instead of `children/child`. It would eliminate the parent/child relationship you were attempting to avoid in the first place.

adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -0,0 +1,106 @@
...
+                    + self.child_padding
+                )
+                sub_view["view"].y = (
+                    int(grid_position_y * self._height / grid_size_y)
+                    + self.child_padding
+                )
+
+                self.append(sub_view["view"])
+
+    def add_sub_view(self, new_view, grid_position, view_grid_size):
```
kattni: I am unclear on where we landed with the use of `view` from your reply in #1. This is still unclear to me what it should mean (beyond the fact that I've used it so I know from context). Everything I'm thinking of is longer though which would be better to avoid. `add_content` maybe?

adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -0,0 +1,106 @@
...
+        """Add a child to the grid.
```
kattni, line 115: This could be clearer. You're adding content to a grid space. I'm struggling to come up with a suggestion for you though.

adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -0,0 +1,106 @@
...
+
+        :param new_view: the child object to add e.g. label, button, etc...
```
kattni: `new_view` is also unclear to me except that it's documented. `cell_object` or since you don't have to type it out in the code, `cell_content_object` perhaps. This is all assuming you decide to go with cell, obviously.

As well, eliminate use of `child` here. I don't think it's needed.

adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -0,0 +1,106 @@
...
+        :param grid_position: where in the grid it should go. Tuple with
+         x,y coordinates in grid cells. e.g. (1,0)
+        :param view_grid_size: the size and shape of cells that the new
+         child should occupy.
+        :return: None"""
+        sub_view_obj = {
+            "view": new_view,
+            "grid_position": grid_position,
+            "view_grid_size": view_grid_size,
```
kattni: This one still sounds to me like I'm looking at the grid size for something.

adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -0,0 +1,106 @@
...
+                self.append(sub_view["view"])
+
+    def add_sub_view(self, new_view, grid_position, view_grid_size):
+        """Add a child to the grid.
+
+        :param new_view: the child object to add e.g. label, button, etc...
+        :param grid_position: where in the grid it should go. Tuple with
+         x,y coordinates in grid cells. e.g. (1,0)
+        :param view_grid_size: the size and shape of cells that the new
+         child should occupy.
```
kattni: Can probably avoid `child` here as well.

adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -0,0 +1,106 @@
...
+import displayio
+
+__version__ = "0.0.0-auto.0"
+__repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout.git"
+
+
+class GridLayout(displayio.Group):
+
+    """
+    A layout that organizes children into a grid table structure.
```
kattni, line 35: Let's move away from `children` here as there are viable options like `cells` etc.

kattni (changes requested):
This is great! Thank you for getting this going.

I have some inline suggestions, and will probably have more, but this is a start. A lot of it relates to the concept of a `view` still being unclear to me. Bear in mind, this is only a set of suggestions, and if you believe we should stick with the current nomenclature, I will support that as long as it's well documented.

The `display_layout_simpletest.py` file should not be empty. I would suggest for now making it identical to the `gridlayout_simpletest` and as we add more to this library, include the features in the simpletest example, which is to say, it would test all the features of the library. You can include a docstring with that explanation at the beginning of the file so you and others who contribute will remember/know to do that.

adafruit_displayio_layout/layouts/grid_layout.py
FoamyGuy: yes! `add_content` is perfect I think. This way the "thing" being added doesn't have to be any more specidic than content, which can work for many types of Layouts and other 'ModifyingWidgets'.

Conceptually speaking view (as I used it here) would basically mean a/any  displayio.Group subclass that has `width` and `height` properties.

So far, in the other thread I think we are leaning toward `Control` as the term used for this concept moving forward, but nothing set in stone yet.

adafruit_displayio_layout/layouts/grid_layout.py
FoamyGuy: I like it, will change references to child / children to cell. Thank you!

adafruit_displayio_layout/layouts/grid_layout.py
FoamyGuy: Once the rest of the changes are in place, I think `cell_size` will make a more clear name for this one.

FoamyGuy:
I will work on these and push changes tomorrow.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: 2020 Diego Elio Pettenò
+#
+# SPDX-License-Identifier: Unlicense
+
+repos:
+-   repo: https://github.com/python/black
+    rev: stable
```
askpatrickw, line 7: This will need to change to `20.8b1`

.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+    hooks:
+    - id: black
+-   repo: https://github.com/fsfe/reuse-tool
+    rev: latest
```
askpatrickw, line 11: This will need to change to `v0.12.1`

askpatrickw (commented):
I noted two changes needed to make pre-commit happy, PR to the cookiecutter for the same fix is pending.

FoamyGuy:
> I noted two changes needed to make pre-commit happy, PR to the cookiecutter for the same fix is pending.

Thank you! I'll get those in tonight along with the other changes.

kattni:
@FoamyGuy Hold off on the pre-commit changes, we're looking into those. Should have an answer soon.

kattni:
@FoamyGuy Thanks for replying! I'm glad my suggestions work for you. Looking forward to the updates! I'll test it once you get it settled.

FoamyGuy:
I think this is all set now, pending the potential pre-commit changes. I did go ahead and put a copy of the grid_layout example into the main simpletest script. We can add to that as we get other layouts and widgets in this library.

@kattni I believe it should be functioning the same as the copy you started with, but would not be an actual drop-in replacement due to the changed names of some functions, arguments, and properties.

examples/displayio_layout_gridlayout_simpletest.py
```diff
@@ -0,0 +1,53 @@
...
+display.show(main_group)
+
+layout = GridLayout(
+    x=10,
+    y=10,
+    width=200,
+    height=100,
+    grid_size=(2, 2),
+    cell_padding=8,
+    max_cells=10,
```
kmatch98: Should read `max_size=10`.

examples/displayio_layout_simpletest.py
```diff
@@ -0,0 +1,55 @@
...
+display.show(main_group)
+
+layout = GridLayout(
+    x=10,
+    y=10,
+    width=200,
+    height=100,
+    grid_size=(2, 2),
+    cell_padding=8,
+    max_cells=10,
```
kmatch98: Should read `max_size=10`.

kmatch98 (commented):
#### Examples
Two of the examples look to be almost the same, or am I missing something?  
- `examples/displayio_layout_gridlayout_simpletest.py`
- `examples/displayio_layout_simpletest.py`

#### Resizing and centering widgets in cells (plus a rabbit hole of `label` and `bitmap_label`)
One thing that I realized is that some Widgets' width and height may have constraints on them.  I suggest that all Widgets be resized using tuples (width, height) so that the Widget can decide on its own aspect ratio.

On a related note, a `label` or `bitmap_label` dimenions cannot actually be modified.  Therefore, if `grid_layout` tries to resize a label, the `label.width` and `label.height` should actually not change.  

That brings me to the next point, that I think `grid_layout` should default to centering the Widget in the cell.

But there are problems in `label` and `bitmap_label` libraries that need to be resolved to allow this.

For the simpletest example, it plots four labels, but they are all in the upper left corner of the cell.  I think the default should be that items inside of a cell should be centered.  In looking deeper at the `label` structure, the `label.width` and `label.width` actually mean something else in the `label` Class.  Instead, currently the `bounding_box` is where the actual pixel dimensions of the `label` are kept.  (Currently `label.width` is the number of characters and `label.height` is the pixel height of the text.)

Also, another related issue that will arise is that `bitmap_label` does not have any `.height` or `.width` property.

I think if we define that Widget's should have properties of `height` and `width`, then we'll have to go change `label` and `bitmap_label` to accommodate this.

Anyway, I think that `grid_layout` placement method should be modified so that it places the x,y locations of the Widget and centers it in the cell.  That will require the `_layout_cells` function to place the x position something like this:
```python
cell["content"].x = (
                    int(grid_position_x * self._width / grid_size_x) + self.cell_padding 
                       + int( (self._width/grid_size_x) - cell["content"].width) /2 )
                )

cell["content"].y = (
                    int(grid_position_y * self._width / grid_size_y) + self.cell_padding 
                       + int( (self._height/grid_size_y) - cell["content"].height) /2 )
                )
```

#### Summary
1.  Update `grid_layout` to center items in cells by checking the item's actual width and height.
2. Update `label` so the `height` and `width` mean what `grid_layout` wants
3. Add `height` and `width` getters/setters to `bitmap_label`

I didn't get around to checking buttons.....

examples/displayio_layout_simpletest.py
FoamyGuy: thank you! fixed with latest commit

examples/displayio_layout_gridlayout_simpletest.py
FoamyGuy: thank you! fixed with latest commit

FoamyGuy:
@kmatch98 Thank you for taking the time to go over this and review. Especially catching the bad arguments in the examples!

It true that `label` and `bitmap_label` don't currently support resizing so currently the grid layout tries and fails to manipulate them like that. I have a very rough cut of my attempt at a solution for this here: https://github.com/FoamyGuy/Adafruit_CircuitPython_Display_Text/blob/dynamic_sizing/adafruit_display_text/static_sized_bitmap_label.py (it could probably use a better name)

I decided to make a new kind of label instead of changing the existing ones. I think the way the existing ones work is nice as well because they can grow to fit whatever text you want to put inside which is desirable sometimes. It could perhaps be modified to work both ways within the same class but the code would get more complex and larger to import. I think right now I'm in favor of leaving the existing ones alone and encouraging people to use this new type of label for best support with grid_layout.

The new one started as a copy of bitmap_label and I changed up the parts necessary for it to have `height` and `width` properties. I intend to add alignment options as well that will allow user code to align the text how they want within the box defined by width and height. This would allow for the centered alignment like you describe, if this new label is used.

However I do still want to have some decent support for the existing bitmap_label and label and more generally speaking widgets that don't allow size manipulation so I do also like your idea of centering widgets within the cell. I'd love to ultimately see a configuration option that allows aligning widgets smaller than the cell to either left, center, or right. I will start with centering for now but look to add that enhancement sometime after this initial version code is in.

With regards to setting size via tuple I'm not opposed, but I do like the ability to change one without worrying about the other that separate properties allows. I'm not sure if I understand if or why the tuple is required for widgets to maintain their own aspect ratio. Couldn't similar logic live in the height and width properties? so whenever user changes one of them it will adjust it's internal "actual drawn size" (which is clamped to the aspect ratio desired) to fit within the new bounds based on whichever property the user changed.

kmatch98:
Here’s some clarification on my proposal to set width and height in a tuple. 

Imagine a widget that likes to be in a ratio of 3:1width to height. 

When it’s first created it is 30 wide, 10 high. 
Now we change to width of 90. 
Since the height is still 10, should it just stay 30 wide and still 10 high?

Assuming that the widget always stays inside the maximum size it can within width height, then it keeps an actual size of 30 x 10. 

Next we say to change height to 30. 

So does it need to remember that it’s actual height and width is currently 30x10 but the “last_requested” width was 90?

I think that will cause way to much trouble to keep some memory of what was requested. 

Instead I propose that you send a tuple of (width, height) and the widget will decide how to best fit it into that maximum boundary. 

In contrast, if we set one at a time I think we can expect to have some weird behavior depending upon the sequence of width and height requests. 

I do think Widgets can keep separate width and height setter functions, but for use with grid_layout I think the each shape should be set with a tuple. 

And even better (if we don’t want to be too strict with the Widget class definition), we can have grid_layout try using the tuple first if the Widget tuple function is available, and if not it can fall back to individual width and height setters. 

Hope I didn’t muddy the waters any worse. 

I really think grid_layout is going to be useful for anyone trying to make a quick GUI.

FoamyGuy:
I do see what you mean about setting the sizes individually causing potential weird behavior.

In the Resizeable button PR I've added a `resize(new_width, new_height)` function that can be used to set both in one action.

In this PR the latest commit makes use of that new resize function if the content object has it, but still falls back to width and height properties if not.

kmatch98:
I tested the `resize(width, height)` function and it works well.  So far, everything works for me, so I suggest merging.

kattni (approved):
I still want to test this on my project, but I don't think merging needs to wait on me. I can file an issue if I find anything!

Issue #3, opened 2021-01-28
Draft proposal of standard parameters and functions for DisplayIO Widgets


kmatch98:
The DisplayIO_Layout library will rely on having several standard parameters for any "widgets" so this function can place them in the desired layout.  Additionally, for any touch-related objects, it is useful to have standard response functions.

The starting point for this proposal was the [Adafruit_CircuitPython_Display_Text library](https://github.com/adafruit/Adafruit_CircuitPython_Display_Text) with the [label item](https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/blob/master/adafruit_display_text/label.py) along with the [Adafruit_CircuitPython_Display_Button library](https://github.com/adafruit/Adafruit_CircuitPython_Display_Button) with the [button item](https://github.com/adafruit/Adafruit_CircuitPython_Display_Button/blob/master/adafruit_button.py).

I'd especially like proposals on the standardized "Response functions" for widgets.

---------

#### Proposed widget parameters and naming:
* Position and sizing
    * `x`, `y` : upper left corner of the widget, in display pixel coordinates (see diagram)
    * `anchor_point` : (a, b) two values between 0 and 1.  Widget placement using `anchor_point` and `anchored_position` should operate the same as display_text.Label (see candy hearts example: https://learn.adafruit.com/circuit-python-tft-gizmo-candy-hearts/how-it-works)
    * `anchored_position` : (C,D) in pixels (see `anchor_point` description above)
    * `width` : in pixels
    * `height` : in pixels
    * `bounding_box` : (x, y, width, height), *getter only
* Touch-related
    * `touch_padding` : in pixels, additional space around the `bounding_box` that accept touch input (alternately, modify `touch_boundary` directly)
    * `touch_boundary` : (x, y, width, height) region that will accept touch input
* Response functions and data
    * `contains(touch_point)` : responds True if touch-point is within the widget’s `touch_boundary`
    * `selected(touch_point)` : widget was just touched, reaction to selection will depend upon the widget’s needs
    * `still_touched(touch_point)` : widget remains touched (***?)
    * `released(touch_point)` : widget is released (***?)
    * `value` : type will depend upon the widget (getter/setter required)
* Widget naming (*** Should each widget have a label that can be optionally displayed and positioned?) 
    * `label` : text label to describe the button
    * `display_label` : Boolean, set `True` to display the name of the widget
    * `label_font` : font 
[...]

PR #5, opened 2021-02-19
Main classes: `Widget` and `Control`

kmatch98:
This PR adds the primary main classes that define the baseline structure of the `Widget` and `Control` classes.

The main discussion related to the structure is located in this [issue](https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/issues/3).


kmatch98:
This is ready fo review. I have simplified [the earlier draft PR](https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/pull/4) by just pulling out the main class definitions for `Widget` and `Control`. I have not included any widget examples here but the draft PR includes several that will eventually each have their own PR. 

Main objective of this review is to get feedback on the class definitions and whether these two classes have the right structure to accommodate an assortment of future graphical widgets, including those with touch response. 

Some lingering questions to spur discussion:
- In the absence of using interrupts, we will rely on a main loop to handle touch events and call the appropriate widgets. In my current widgets, and widget actions are blocking (such as animation of a switch moving).  Do we instead want “non-blocking” widget response functions and then a separate “redraw” widget function?  If so, the main loop will need to call the “redraw” function on each widget in case any redrawing is required (such as during animation) while the main loop can still watch for touch inputs. 
- Should each widget have a built-in option for a label?  Currently I kept a place holder `name` for this. 
        - Benefits: new users could use one command to add a widget and also label it.  
        - Cons: Not all widgets have an obvious place to put a label, and a separate Widget_Label could do the same job. 
- Are there other class variables or response functions that should be added to cover a broader set of use cases?

Thanks for your input!

adafruit_displayio_layout/widgets/widget.py
```diff
@@ -0,0 +1,164 @@
...
+class Widget(displayio.Group):
+    """
+    A Widget class definition for graphical display elements.
+
+    :param int x: pixel position
+    :param int y: pixel position
+    :param int width: width of the switch in pixels, set to ``None`` to auto-size
+     relative to the height
+    :param int height: height of the switch in pixels
+    :param str name: name of the switch
```
FoamyGuy: A few of these parameter docstrings reference `switch` instead of `widget` 

height, width, name, anchor_point

adafruit_displayio_layout/widgets/widget.py
```diff
@@ -0,0 +1,164 @@
...
+    :param float anchor_point: (X,Y) values from 0.0 to 1.0 to define the anchor
+     point relative to the switch bounding box
+    :param int anchored_position: (x,y) pixel value for the location
+     of the anchor_point
+
+    """
+
+    def __init__(
+        self,
+        width=None,
```
FoamyGuy, line 172: In the docstring above `x` and `y` are listed as the first two arguments, but they aren't here inside of __init__ parameters.

Is that because they get brought in with **kwargs below? If so I wonder if it would break anything if we added them explicitly I think for the sake of reading the code and understanding it would be nice if they were listed here if possible.

adafruit_displayio_layout/widgets/widget.py
```diff
@@ -0,0 +1,164 @@
...
+        height=None,
+        name="",
+        anchor_point=None,
+        anchored_position=None,
+        bounding_box=None,  # pixel extent of the widget [x0, y0, width, height]
```
FoamyGuy: Would anything break if someone passed in a `bounding_box` that was different than the x, y, width, and height that they also pass in?

FoamyGuy (commented):
These are looking pretty good to me. I noticed a few docstrings that need tweaked, and I had a few other questions.

I can share my thoughts on the questions you asked as well:
Q1: I think I am in favor of making them blocking and not having the user rely on calling an update or animate function in their main loop. The animation should be taking _too_ long (though that depends a lot on the use case as well). And I think it's better to block so that it can complete uninterrupted. If user needs no block they could set animation time to 0.

Q2: I think I lean towards no `name` by default even though it would make some use cases a little easier. It's not a super strong opinion though, I'm open to other thoughts on this. If we do build it in to widget I think it would be best if it can be done in such a way that if the user is not using it then they can still use their widget without needing display_text library on their device (assuming they aren't otherwise using it.)

Q3: for the short / medium term I think this set that you have is good. If anything I might be in favor of removing a few until we get a more fleshed out event system in place. I'd maybe get rid of `gesture_response`, `released`, and `still_touched`. They could get added back in once we have something specific that is using them.

It's definitely a larger discussion than for this PR, but in the longer term I wonder if there is any opportunity for a core module to interact with a touch screen "in the background" and cache events for us waiting until the user code checks for them. Similar to how displayio renders in the background, and gamepad manages button state in the background.

adafruit_displayio_layout/widgets/widget.py
kmatch98: Thanks for the suggestion, I struggled with whether I should explicitly call them out here or if someone should have to go up the class tree to find them.  

I agree with you to include x,y in the `__init__` definition. So, I added them back in here.  

Also I added scale, along with a warning, that if scale > 1 that you have to take care of scaling the `touch_boundary`.

adafruit_displayio_layout/widgets/widget.py
kmatch98: Thanks for the feedback.  I decided to get rid of `bounding_box` as an input parameter.  It's more confusing than useful.

kmatch98:
I made major updates to the documentation and was able to build them on my machine.

However, the checks failed here due to Pylint.  It's complaining about repeated code use in the examples folder.  Strange.  I suspect this is due to some unknown changes in the pylint settings.  @kattni will look into this and see if there is a way around it.

Please wait until that is resolved before spending any time on this, since the documentation is probably the biggest change added in the latest commits.  Let's give it a few days to get resolved and then can review once the docs are built.

kmatch98:
Update on Q1 above. 

There are asynchronous (async) features that have been added recently:
- https://github.com/WarriorOfWire/tasko
- https://github.com/adafruit/circuitpython/pull/3540

This can come in handy for the widget redraw functions and the touch response handler. 
 
But for now let’s keep the redraw code as blocking, and explore this async capability in the future.

kmatch98:
Thanks to the recent pylint fixes, this is now passing.  

@FoamyGuy I'd appreciate if you would take a quick look to verify I addressed your comments.

FoamyGuy (approved):
These are looking great to me after the latest changes. Thank you! I think this is ready to merge any time. Let me know if there is anything you know of we need to hold back for @kmatch98. If not i'll be back this evening to merge this.

PR #6, opened 2021-02-24
Feature/fixup pylint example errors

hugodahl:
This add calls to `pylint` through the pre-commit hooks, instead of the GitHub workflow.

This is a draft PR until  the GitHub workflow actions in `build.yml` are also updated.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,3 +17,11 @@ repos:
     -   id: check-yaml
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
+-   repo: local
+    hooks:
+    -   id: pylint_examples
+        name: pylint-examples
+        entry: /usr/bin/env bash -c
+        args: ['([[ ! -d "examples" ]] || for example in $(find . -path "./examples/*.py"); do pylint --disable=missing-docstring,invalid-name,bad-whitespace $example; done)']
+        language: system
+        description: Run example-specific pylint rules
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,9 +19,15 @@ repos:
     -   id: trailing-whitespace
 -   repo: local
     hooks:
+    -   id: "pylint-library"
+        name: "pylint-library"
+        entry: /usr/bin/env bash -c
+        args: ['pylint $( find . -path "./adafruit*.py" )']
+        language: system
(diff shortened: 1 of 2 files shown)


hugodahl:
For now, both calls to `pylint` are done through shell commands, so that the changes could be made as quickly as possible to make it work.

Ultimately, at the very least, the checks for library files, `adafruit_*.py`, should run using the existing pre-commit hook for pylint from the repo at "https://github.com/pycqa/pylint". See [issue #104](https://github.com/adafruit/cookiecutter-adafruit-circuitpython/issues/104) in the [Cookiecutter](https://github.com/adafruit/cookiecutter-adafruit-circuitpython/) repository.

gamblor21 (approved):
Checked the actions output. pylint ran as part of pre-commit and not its own action.

.pre-commit-config.yaml
```diff
@@ -17,3 +17,17 @@ repos:
     -   id: check-yaml
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
+-   repo: local
```
kattni, line 20: Please explain how this works. I didn't look into it too deeply, but my first impression was that `local` meant local to your machine. Will this still work in the library when someone PRs to it?

kattni (changes requested):
Thank you so much for doing this! Let's hold off on this PR though, and get it into `cookiecutter` first if this is how it should be done. We'll do a patch of all the libraries after that. Please see my question below.

.pre-commit-config.yaml
hugodahl: Right, it means that it's not pulling the code that pre-commit runs from another repo, but rather, it's executing things locally.

Since part of the GitHub workflow is to install pylint, and that happens before pre-commit runs when someone creates a PR against a GitHub repository, we will be assured that pylint is installed.

The downside is that pylint isn't automatically installed as it would with pre-commit. If we want, we _could_ add that to the custom hook, however, I know we'd rather not pollute peoples' environments by installing things willy-nilly.

hugodahl:
Closing this issue since the root fix is addressed with [PR #109](https://github.com/adafruit/cookiecutter-adafruit-circuitpython/pull/109) in the [Cookiecutter](https://github.com/adafruit/cookiecutter-adafruit-circuitpython/) repo.

PR #10, opened 2021-03-10
rearranged documentation

kmatch98:
This rearranges the `switch_round` documentation so that the parameters show up at the top in the ReadTheDocs page.

I think this is cleaner so that the parameters show up first when you pull up the RTD page for this widget.


FoamyGuy (approved):
I tested this out building locally with sphinx. 

Agreed it's nicer with the parameters listed first. Changes look good to me. 

Thank you @kmatch98 for this really nice docs page, and the improvement to it with this PR!

PR #11, opened 2021-03-12
Optional position in resize

FoamyGuy:
This adds an optional argument for `position` to the resize function in Widget. 

subclasses can make use of this argument to do things that require knowing the position such as updating the touch_boundry to reflect the new size and location of a widget after it's been moved.

Using these changes and GridLayout I was able to make a grid of IconWidgets that are all individually touch capable.

image

### DIFF
--- a/adafruit_displayio_layout/layouts/grid_layout.py
+++ b/adafruit_displayio_layout/layouts/grid_layout.py
@@ -73,9 +73,10 @@ class GridLayout(displayio.Group):
                 button_size_y = cell["cell_size"][1]
 
                 new_position = (
-                    int(grid_position_x * self._width / grid_size_x) + self.cell_padding,
+                    int(grid_position_x * self._width / grid_size_x)
+                    + self.cell_padding,
                     int(grid_position_y * self._height / grid_size_y)
-                    + self.cell_padding
+                    + self.cell_padding,
                 )
 
                 print(
@@ -95,7 +96,7 @@ class GridLayout(displayio.Group):
                             int(button_size_y * self._height / grid_size_y)
                             - 2 * self.cell_padding
                         ),
-                        new_position
+                        new_position,
                     )
                 else:
                     # try width and height properties.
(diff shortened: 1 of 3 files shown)


kmatch98:
@foamyguy I don’t understand the need for sending the position when resizing a Widget. When a widget gets resized it should automatically update its touch boundary to reflect the new size. A widget call always get its own new position from self.x and set.y. 

However I do understand the need for grid layout having a feature like `anchor_point` to allow centering or justifying widgets inside the grid. Can this be solved using an `anchor_point` input to grid_layout?

I think I may be missing what your getting at here. 

In SwitchRound here’s how I adjust for the local position when calling the superclass `Control.contains()`:
```python
def contains(self, touch_point):  # overrides, then calls Control.contains(x,y)
        """Checks if the Widget was touched.  Returns True if the touch_point
        is within the Control's touch_boundary.
        :param touch_point: x,y location of the screen, in absolute display coordinates.
        :return: Boolean
        """
        touch_x = (
            touch_point[0] - self.x
        )  # adjust touch position for the local position
        touch_y = touch_point[1] - self.y

        return super().contains((touch_x, touch_y, 0))
```

FoamyGuy:
@kmatch98 the code that uses it is here currently: https://github.com/FoamyGuy/Touch_Deck_Working_Files/blob/main/touch_deck_code.py

In particular it's used inside of IconWidget in order to update the touch_boundry to the new size and positioned that was set by the grid_layout.

Nice idea to override contains to adjust the coordinates. That is pretty much what I need I think. I'll give that a try and close this if it works how I'm thinking.

FoamyGuy:
@kmatch98 if you find yourself curious this was the code at the time I made the change: https://github.com/FoamyGuy/Touch_Deck_Working_Files/commit/f2171c260078aa319d288b6582e6d65ec44667f6#diff-538a96780fd1dad5b3b939f4b6280ae5e39d202c22811192e647c73cbca3a677

But I have successfully changed it over to use the same technique you mentioned here from the switch.

Thank you for the pointer!

PR #13, opened 2021-03-14
Dial gauge

kmatch98:
dial

Adds a dial gauge to the CircuitPython Widget family.  Includes some brief documentation with images showing some of the main features.


docs/api.rst
```diff
@@ -23,6 +23,14 @@
...
+   :members:
+   :member-order: bysource
+   :inherited-members:
+
+.. inheritance-diagram:: adafruit_displayio_layout.widgets.dial
+
 .. automodule:: adafruit_displayio_layout.widgets.icon_widget
    :members:
    :member-order: bysource
+
```
FoamyGuy: I think it wants one of the newlines at the end of this file deleted. It's picky about having exactly 1 empty newline at the end. I made this change locally and it seems to pass pre-commit after that.

FoamyGuy (commented):
This looks awesome! thanks for adding it. I will take a closer look and test it out tomorrow.

docs/api.rst
kmatch98: Thanks, for the note on this.  I updated and resubmitted.

examples/displayio_layout_dial_simpletest.py
```diff
@@ -0,0 +1,61 @@
...
+
+# Define the minimum and maximum values for the dial
+minimum_value = 0
+maximum_value = 100
+
+# Create a Dial widget
+my_dial = Dial(
+    x=20,  # set x-position of the dial inside of my_group
+    y=20,  # set y-position of the dial inside of my_group
+    width=180,  # requested width of the di
```
FoamyGuy: this comment got cut off I think. should be `diameter`

adafruit_displayio_layout/widgets/dial.py
```diff
@@ -0,0 +1,1096 @@
...
+        anchored_position=None,
+        tick_color=0xFFFFFF,
+        major_ticks=5,  # if int, the total number of major ticks
+        major_tick_stroke=3,
+        major_tick_length=10,
+        major_tick_labels=("0", "25", "50", "75", "100"),
+        minor_ticks=5,  # if int, the number of minor ticks per major tick
+        minor_tick_stroke=1,
+        minor_tick_length=5,
+        tick_label_font=None,
```
FoamyGuy, line 189: I think this should default to `terminalio.FONT` so if user does not pass anything they will still be able to make a dial.

adafruit_displayio_layout/widgets/dial.py
```diff
@@ -0,0 +1,1096 @@
...
+       and major and minor tick marks.
+    """
+
+    # The dial is a subclass of Group->Widget.
+
+    def __init__(
+        self,
+        width=100,
+        height=100,
+        padding=0,  # keepout amount around border, in pixels
```
FoamyGuy: I think this should default to something like `12` or so. At `0` it seems to cut off parts of the labels when a dial is created with minimal arguments passed. At `12` the minimal argument dial is looking nice.

adafruit_displayio_layout/widgets/dial.py
```diff
@@ -0,0 +1,1096 @@
...
+    :param int x: pixel position
+    :param int y: pixel position
+
+    :param int width: requested width in pixels
+    :param int height: requested height in pixels
+    :param int padding: keepout padding amount around the border, in pixels
+
+    :param float sweep_angle: dial rotation, in degrees, maximum value is 360 degrees
+    :param float start_angle: starting angle, in degrees.  Defaults
+     to `None` for symmetry along vertical axis
```
FoamyGuy: I think we should mention in these `angle` parameter docstrings that 0 degrees is north or up or 12 o'clock (I'm not sure which term is best to convey the idea). Also a note that negative values will move to the left from there counter-clockwise.

FoamyGuy (commented):
This is really amazing! thanks for working on this @kmatch98 

I have a few minor suggestions, but overall this is looking pretty good to me.

I tested the new example, and a few variations on it with: 
```
Adafruit CircuitPython 6.2.0-beta.3-36-g103632f44 on 2021-03-07; Adafruit PyPortal with samd51j20
```
I also built locally and checked out the new sections of the docs pages.

adafruit_displayio_layout/widgets/dial.py
kmatch98: Updated per your suggestion.

adafruit_displayio_layout/widgets/dial.py
kmatch98: Updated per your suggestions, both for `tick_label_font` and `value_font`.

adafruit_displayio_layout/widgets/dial.py
kmatch98: Updated docstring per your suggstion.

examples/displayio_layout_dial_simpletest.py
kmatch98: Corrected, good catch.

adafruit_displayio_layout/widgets/dial.py
```diff
@@ -135,6 +136,22 @@ class Dial(Widget):
...
+
+    .. figure:: dial_variables_clip_needle.png
+       :scale: 35 %
+       :figwidth: 70%
+       :align: center
+       :alt: Diagram showing the impact of ``clip_needle`` Boolean value.
+
+       Diagram showing the impact of the ``clip_needle`` input parameter,
+       with the dial's boundary shown. For ``sweep_angle`` values less than
+       180 degrees, the needlecan protrude a long way from the dial ticks. By
```
FoamyGuy: minor typo here, need space between "needle can"

adafruit_displayio_layout/widgets/dial.py
```diff
@@ -135,6 +136,22 @@ class Dial(Widget):
 
        Diagram showing the various parameters for setting the dial labels
        and major and minor tick marks.
+
+    .. figure:: dial_variables_clip_needle.png
```
FoamyGuy, line 141: I'm not seeing these images in my docs page when I try to build it locally.  Do these show up for you? I could be doing something wrong, I don't think I've ever built with images before.

FoamyGuy (commented):
The code in this is looking good to me now. I noticed one small typo, and have a question about the docs pages.

adafruit_displayio_layout/widgets/dial.py
kmatch98: Great catch, fixed.

adafruit_displayio_layout/widgets/dial.py
kmatch98: Yes, these do show up on my builds locally.  Can you see the graphic for the widget (see clip below)?  The widget graphic definitely made it ok onto the official released ReadTheDocs.  I'm a real novice at Sphinx and ReadTheDocs, so maybe you can trace that better than I ever could.

image

FoamyGuy: Ah, yeah I see that in the published docs. But not my local ones. I must be missing something locally.

kmatch98: Here’s a comment about the directory structure that is expected to include figures (absolute vs relative). If it builds but does not show the graphics maybe it’s not finding the image files?

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#images

FoamyGuy (approved):
I re-tested the latest version with a PyPortal. It's working well.

This looks good to me. Thanks again for all of your work on this @kmatch98

ladyada:
so cute!

PR #15, opened 2021-03-15
icon_widget: Add animation

kmatch98:
This adds a zoom animation when the widget is `selected` and `released`.  

icon_widget animation

I added this feature to @FoamyGuy 's touch-deck code, with the example code here:
https://gist.github.com/kmatch98/aeae5c3b34754050037a294348416c52

To reduce memory usage, I used smaller icons with reduced colors compared to their original `touch_deck_icons` files.

icon_animate_palette

I updated this PR to include an option for `_pixelbuf.colorwheel` palette animation for `unselected` icons.  This adds several input parameters to control the palette animations, including an optional list of palette indices that are left untouched.

### DIFF
--- a/adafruit_displayio_layout/widgets/icon_widget.py
+++ b/adafruit_displayio_layout/widgets/icon_widget.py
@@ -25,6 +25,7 @@ import gc
 import time
 import terminalio
 import bitmaptools
+from _pixelbuf import colorwheel
 from displayio import TileGrid, Bitmap, Palette
 import adafruit_imageload
 from adafruit_display_text import bitmap_label
@@ -62,6 +63,17 @@ class IconWidget(Widget, Control):
     :param int max_size: (Optional) this will get passed through to the
      displayio.Group constructor. If omitted we default to
      grid_size width * grid_size height to make room for all (1, 1) sized cells.
+    :param int wheel_initial_value: When using palette animation, this is the initial value
+     of the colorwheel parameter used with ``_pixelbuf.colorwheel``
+    :param int wheel_increment: To add palette animation, set this to the value of
+     how much you want the ``_pixelbuf.colorwheel`` function to increment each time that
+     ``unselected`` is called (default: 0 for no palette animation)
+    :param int wheel_grading: This is the step sized used when calling colorwheel for
+     each color index in the palette (default: 5), basically it's how far apart each
+     color in the palette will be set. Use a low value if you want each color to be
+     close to each other or a high value to spread out into a wider range of colors.
+    :param int palette_skip_indices: integer or list of integers with the palette
+     indices that should not be changed when using the palette animations (default: None)
 
     """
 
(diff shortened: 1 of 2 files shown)


adafruit_displayio_layout/widgets/icon_widget.py
```diff
@@ -50,28 +63,92 @@ class IconWidget(Widget, Control):
...
+    def __init__(
+        self,
+        display,
+        label_text,
+        icon,
+        max_scale=1.0,
+        max_angle=8,
+        animation_time=0.0,
+        wheel_initial_value=1,  # initial wheel color value
+        wheel_increment=0,  # how much the wheel
```
FoamyGuy: missing the end of the sentence in this docstring.

adafruit_displayio_layout/widgets/icon_widget.py
```diff
@@ -89,3 +166,125 @@ def contains(self, touch_point):  # overrides, then calls Control.contains(x,y)
         touch_y = touch_point[1] - self.y
 
         return super().contains((touch_x, touch_y, 0))
+
+    def selected(self, touch_point):
```
FoamyGuy: With regards to naming: If we use "selected" and "unselected" I think they should be roughly the opposite of each other. So if selected is going to do the zoom in, then I think unselected should do the zoom out.

Truthfully I lean toward wanting some different words.  `selected` is kind of nice since it's in the Button API already. But I think the name can be somewhat confusing because "selected" sometimes mean "has focused" not "being clicked". 

I think we should use names that describe the animation. That way folks can tell what they do at a glance and call the ones they want to use when appropriate.

So for the animations in this IconWidget so far I would recomend: `zoom_animation()`, `zoom_out_animation()`, `colorcycle_animation()`

adafruit_displayio_layout/widgets/icon_widget.py
```diff
@@ -50,28 +63,92 @@ class IconWidget(Widget, Control):
...
+        label_text,
+        icon,
+        max_scale=1.0,
+        max_angle=8,
+        animation_time=0.0,
+        wheel_initial_value=1,  # initial wheel color value
+        wheel_increment=0,  # how much the wheel
+        wheel_grading=5,  # sets the colorwheel distance between palette colors
+        palette_skip_indices=None,  # single value or list of palette color indices to
+        # remain constant during animations
```
FoamyGuy: this comment line feels like it should be indented over underneath the line above. But since black and pylint both ran and didn't complain maybe my feeling is off

FoamyGuy (commented):
Thanks for the work on this @kmatch98. This is some really nice functionality that can polish up an interface nicely.

I found one docstring needing finished. And I have a few suggestions for naming. But beyond those this is looking good to me.

kmatch98:
I will cancel this PR and will restart a new one due to conflict with some other changes.

PR #17, opened 2021-03-17
on_disk argument for IconWidget to use OnDiskBitmap

FoamyGuy:
This adds on `on_disk` argument to the IconWidget constructor. If set to `True` it will use OnDiskBitmap instead of imageload which allows it to use less memory and seems to draw a bit quicker.

I've defaulted it to `False`  because imageload allows for animating the icons, and I am not sure if that is possible with OnDiskBitmap. So user can choose which is more important to them for their specific device and project.

I tested this using the most recent version of the touch_deck_code.py from here: https://github.com/FoamyGuy/Touch_Deck_Working_Files on a Feather RP2040

### DIFF
--- a/adafruit_displayio_layout/widgets/icon_widget.py
+++ b/adafruit_displayio_layout/widgets/icon_widget.py
@@ -24,7 +24,7 @@ Implementation Notes
 
 
 import terminalio
-from displayio import TileGrid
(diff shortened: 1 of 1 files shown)


jepler (changes requested):
Thank you.  Please add the new constructor argument to the class docstring!

jepler (approved):
I see it now

PR #18, opened 2021-03-17
Add zoom animation to icon_widget presses

kmatch98:
111096490-a12bbf80-850d-11eb-8ce5-561e5f8251b2

Adds animation to the icon_widget, including `zoom_animation` and `zoom_out_animation`.

Now accommodates the OnDiskBitmap loading to save RAM.  Utilizes an class variable as a bitmap buffer.  Current definition of the class variables gives room for an 80x80 icon bitmap with a maximum zoom of 1.5x, and color depth up to 512.

My updated test files for the Touch Deck are here: https://github.com/kmatch98/Touch_Deck_Working_Files

(Note: This PR does not include the palette colorwheel demo when "unpressed".  I will need additional time to understand how the combination of OnDiskBitmap and `ColorConverter()` works so that we can coopt it for this rainbow-y purpose.)


ladyada:
haha this has strong ps1 vibes

kmatch98:
@ladyada If we can get a color wheel into ColorConverter/OnDiskBitmap, maybe we upgrade to a PS2 vibe? 

https://cdn.discordapp.com/attachments/682423270083002468/821526384677617704/video0.mov

ladyada:
ya gotta have it make the crash bandicoot sfx 😂

adafruit_displayio_layout/widgets/icon_widget.py
```diff
@@ -52,19 +64,67 @@ class IconWidget(Widget, Control):
...
+        print("kwargs: {}".format(kwargs))
+
+        super().__init__(**kwargs)  # initialize superclasses
+        super(Control, self).__init__()
+
+        self.display = display
+        self._icon = icon
 
         if on_disk:
+            print("on_disk")
```
FoamyGuy: lets remove this debugging print.

adafruit_displayio_layout/widgets/icon_widget.py
```diff
@@ -52,19 +64,67 @@ class IconWidget(Widget, Control):
...
+        self,
+        display,
+        label_text,
+        icon,
+        on_disk=False,
+        max_scale=1.0,
+        max_angle=8,
+        animation_time=0.0,
+        wheel_initial_value=1,  # initial wheel color value
+        wheel_increment=0,  # how much the wheel
```
FoamyGuy: I think the end of the sentence in this argument comment got cut off perhaps.

adafruit_displayio_layout/widgets/icon_widget.py
```diff
@@ -52,19 +64,67 @@ class IconWidget(Widget, Control):
...
+        wheel_grading=5,  # sets the colorwheel distance between palette colors
+        palette_skip_indices=None,  # single value or list of palette color indices to
+        # remain constant during animations
```
FoamyGuy: this comment line feels like it should be indented over to the right underneath the comment from the line above to me.

But black and pylint both ran without complaining so perhaps my feeling is off on this one.

adafruit_displayio_layout/widgets/icon_widget.py
```diff
@@ -52,19 +64,67 @@ class IconWidget(Widget, Control):
     :param int max_size: (Optional) this will get passed through to the
      displayio.Group constructor. If omitted we default to
      grid_size width * grid_size height to make room for all (1, 1) sized cells.
+    :param int wheel_initial_value: When using palette animation, this is the initial value
+     of the colorwheel parameter used with ``_pixelbuf.colorwheel``
+    :param int wheel_increment: To add palette animation, set this to the value of
+     how much you want the ``_pixelbuf.colorwheel`` function to increment each time that
+     ``unselected`` is called (default: 0 for no palette animation)
```
FoamyGuy: I think `unselected` was removed in the latest changes.

FoamyGuy (commented):
This is looking pretty good to me. Thanks for all of your work on it @kmatch98. 

I found a few small comment and other minor changes, but this looks ready to merge after that to me. I'll come back later tonight to test it out on a device as well. I haven't run the latest version yet.

I could have sworn I submitted this review last night but I must have forgotten to click the final button.

adafruit_displayio_layout/widgets/icon_widget.py
kmatch98: Removed.

adafruit_displayio_layout/widgets/icon_widget.py
kmatch98: Pushed wrong commit.  The updated commit is now the correct one and this section is deleted.

kmatch98:
Argh!  I pushed the wrong branch last night.  

This commit has the version with the following removed: 
 -  "unselected" function
 -  color_wheel animation.  

Ok, I updated the docstrings and excess print statements and repushed.  Sorry for the confusion on this!  Thanks for your support and patience on this one.

adafruit_displayio_layout/widgets/icon_widget.py
```diff
@@ -52,19 +66,47 @@ class IconWidget(Widget, Control):
...
+    _zoom_bitmap = Bitmap(
+        round(_max_scale * _max_pixels),
+        round(_max_scale * _max_pixels),
+        _max_color_depth,
+    )
+    _zoom_palette = Palette(_max_color_depth)
+
+    def __init__(
+        self,
+        display,
```
FoamyGuy: Can we make `display` a keyword arg and add checks where it is used for None. I'm curious if the animation would still work if the user had auto_show enabled on the display object. If so it would be nice to allow users to skip this argument if they want to.

FoamyGuy (commented):
I have a few more suggestions after trying this out: 

It seems like some images are not supported by imageload. I got this error:
```
  File "code.py", line 190, in <module>
  File "/lib/adafruit_displayio_layout/widgets/icon_widget.py", line 240, in zoom_out_animation
  File "adafruit_imageload/__init__.py", line 57, in load
  File "adafruit_imageload/__init__.py", line 48, in load
  File "adafruit_imageload/bmp/__init__.py", line 50, in load
NotImplementedError: True color BMP unsupported
```
I think we should catch this error and just warn the user that the image they have does not support animation.

kmatch98:
Thinking about this a bit more I think it will be better to refactor the animation into a subclass “AnimatedIcon”. There is a lot of baggage that will be unnecessary if someone wants a simple icon. 

I’ll give it another shot at reorganizing it.

adafruit_displayio_layout/widgets/icon_widget.py
kmatch98: Ok, I updated the structure so that the `IconAnimated` is a subclass of` IconWidget`.

I rearranged so that the display is set in a Class initialization method.

kmatch98:
I refactored this into a subclass of `IconWidget` called `IconAnimated`.  This way `IconWidget` can be used without the overhead of the animation buffers. 

The `IconAnimated` subclass has an initialization function `IconAnimated.init_class` where the display and other parameters are passed to initialize the bitmap and palette buffers.  

**Important: The `init_class` function must be called before any `IconAnimated` are instanced.**

Only one icon can be animated at a time.  So, all `IconAnimated` instances share this same zoom bitmap and palette buffers that are stored in the Class variables.  This buffer is used to prevent memory segmentation and memory allocation failures that can be induced by repeated instancing and `del` of large bitmaps.

[The demo version of the touch_deck using `IconAnimated` is located in this gist](https://gist.github.com/kmatch98/40ad578384111cd5e6a63e4a3f36b0da).

FoamyGuy:
@kmatch98 I like the way you refactored this into a subclass. Nice way to add functionality for those that want it, but allow those that don't to skip it.

I'm working on testing this out and I'm getting this error when I attempt to do the class initialization:
```
  File "code.py", line 20, in <module>
TypeError: function missing required positional argument #3
```

My line 20 is this initialization call:
```
IconAnimated(display=display, max_scale=1.5, max_size=(80,80), max_color_depth=256)
```

I'm not sure if I've ever used init_class before, perhaps I am misunderstanding how that works?

adafruit_displayio_layout/widgets/icon_animated.py
```diff
@@ -0,0 +1,243 @@
...
+        cls, display=None, max_scale=1.5, max_size=(80, 80), max_color_depth=512
+    ):
+        """
+        Initializes the IconAnimated Class variables, including preallocating memory
+        buffers for the icon zoom bitmap and icon palette.
+
+        .. Note::  The `init_class` class function should be called before instancing any
+                IconAnimated widgets. Usage example:
+                ``IconAnimated(display=board.DISPLAY, max_scale=1.5,
+                max_size=(80,80), max_color_depth=256)``
```
FoamyGuy: Ah, I figured out the proper syntax. 

I think this usage example needs the `init_class()` added to it.
```
IconAnimated.init_class(display=board.DISPLAY, max_scale=1.5, max_size=(80,80), max_color_depth=256)
```

kmatch98: Yikes, sorry for the error in the docs, that must have been really confusing.  I appreciate that you caught this since that's the important part!

kmatch98:
Thanks for the great catch on the documentation, that was a big one.  I updated and it's ready for your further review.  Thanks again for checking this out and helping improve it to where it is now.

FoamyGuy:
I think that we may have some trouble with anochored positioning in IconWidget. While working on setting up a test for this I placed a single IconAnimated with anchored positioning and noticed that the `touch_boundary` first two items do get updated to the new location on the screen. Which then makes the adjustments done inside of the overridden `contains` method unnecessarily shift the position back toward the origin.

If I position it with `x` and `y` the touch boundary does not get updated and therefore the contains point does register correctly.

kmatch98:
I think this also affects the main branch, but it's ok to discuss here since it will need to be merged into this PR too.

I think the issue can be setting `touch_boundary` in local coordinates.  Then when the  `contains` method is called to check the touches, the IconWidget.contains() method updates the coordinates to local coordinates before  sending to `Control.contains()` which operations on local coordinates.

Right now the current definition of `touch_boundary` uses display coordinates for the upper left corner:
```python
        self.touch_boundary = (
            self.x,
            self.y,
            image.width,
            image.height + _label.bounding_box[3],
        )
```

I think this should be updated to only local coordinates:
```python
        self.touch_boundary = (
            0, # local coordinates
            0, # local coordinates
            image.width,
            image.height + _label.bounding_box[3],
        )
```

I'll push a PR on this and you can evaluate.  Here is the PR: https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/pull/24

kmatch98:
This IconAnimated subclass references the IconWidget superclass for setting the `touch_boundary`, so the PR mentioned above should solve it for this subclass also. 

So, this PR should be ok for `touch_boundary`. Yay for inheritance!

kmatch98:
I just realized I missed this issue:  https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/pull/18#pullrequestreview-614885540

FoamyGuy:
I should be able to give this one a try again tonight. My last few attempts led into finding issues unrelated to this PR with imageload and IconWidget.

I did manage to get a test in last night of the IconAnimated but the animation was looking odd to me. Like the animated image got layered on top of the "real" image but the "real" image wasn't hidden, so during the animation both were visible, albeit with the top one mostly obstructing the bottom one.

I want to  reset my libraries and CircuitPython version to make sure I have a clean environment for testing. I am somewhat suspicious because I've been testing out various different PRs in the core, imageload, and FeatherWing libraries with it. Some of my libraries are modified from their current state. Need to get back to a clean slate.

kmatch98:
I updated this to better handle possible error conditions, including when the `init_class` settings cannot accommodate the requested icon graphics, including requiring that indexed BMP files are required.  

Also, I modified the input parameter names to clarify the difference between the Class variables and the Instance variables.  Please note that the key **instance** inputs have changed to `scale` and `angle`.
The class `IconAnimated.init_class` inputs have changed to `max_icon_size` to differentiate from `max_size` (the Group's max number of elements).  

I also added an example simpletest along with the "play" button icon.

Here are some other icon files:
[icons.zip](https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/files/6184187/icons.zip)

The `truecolor` file should trigger a warning that the animation is cancelled.

FoamyGuy (approved):
I successfully tested the newest version with:
```
Adafruit CircuitPython 6.2.0-beta.4 on 2021-03-18; Adafruit PyPortal with samd51j20
```
Everything is working well.

This looks good to me.

Thanks for all of your work on this @kmatch98! This is a really great addition.

PR #20, opened 2021-03-19
fix setup_py to include the modules

### LINKED ISSUE #14: pip install is missing files
When you install the library via pip it's missing many of the library code files.

I think we need to include more entries in the list in setup.py like this:
```python
packages=[
        "adafruit_displayio_layout",
        "adafruit_displayio_layout.layouts",
        "adafruit_displayio_layout.widgets",
    ],
```
I tested this approach locally by using `pip install .` and it does seem to allow the library to install correctly.

FoamyGuy:
resolves #14

Added the modules into the list in `setup.py` so that they get properly installed via pip.

Tested by running 

```
pip install .
```
In the root of the repo.


kmatch98:
Looks good to me.

PR #21, opened 2021-03-20
remove debug print

FoamyGuy:
There was one debug print leftover the original development in grid_layout. This PR removes it.


kmatch98:
Looks good to me!

PR #24, opened 2021-03-22
update IconWidget touch_boundary to local coordinates

kmatch98:
This corrects the issue observed in the comment: https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/pull/18#issuecomment-803690068

This sets the `touch_boundary` to use local coordinates, rather than global display coordinates.  The `IconWidget.contains()` function then corrects for global display positioning when calling the `Control.contains()` function with local coordinates.

### DIFF
--- a/adafruit_displayio_layout/widgets/icon_widget.py
+++ b/adafruit_displayio_layout/widgets/icon_widget.py
@@ -75,8 +75,8 @@ class IconWidget(Widget, Control):
         )
         self.append(_label)
         self.touch_boundary = (
-            self.x,
-            self.y,
+            0,
+            0,
             image.width,
             image.height + _label.bounding_box[3],
         )


kmatch98:
This update makes the IconWidget’s `touch_boundary` conform to the proposed Widget coordinate system:

https://circuitpython.readthedocs.io/projects/displayio-layout/en/latest/api.html#adafruit_displayio_layout.widgets.widget.Widget

FoamyGuy (approved):
The solution does make sense to me and code looks good.

I tested this successfully with
```
Adafruit CircuitPython 6.2.0-beta.4-43-g7229fe631-dirty on 2021-03-21; Adafruit Feather RP2040 with rp2040
```
Confirmed it does resolve the touch location issue. Both x,y positioning and anchored point positioning now have the same behavior.

Thank you @kmatch98 for fixing this touch boundary that I slipped up on.

PR #25, opened 2021-03-22
Include examples in the API documentation

jposada202020:
To include some examples in the API documentation. They are missing in readthedocs


kmatch98:
This is cool, I learned something today.  I built the docs on my local machine and they look good to me.  Thanks for this update!

jposada202020:
@kmatch98 thanks for the review.

PR #26, opened 2021-03-23
Add limit_rotation option to constrain needle rotation

### LINKED ISSUE #23: Dial gauge can go past limits.
Currently the dial gauge `value` can be set outside the range that is displayed. This causes the needle to go outside the range of motion defined by the `sweep_angle`  

Recommend that the needle position be limited to the specified range.  But also recommend to allow the `value` to be anything in case the user needs the dial to retain this set value, even if outside the min and max values. 

So, only clip the needle position but allow value to be anything. 

(Possible subclass extension: adding a “clip” indicator to a gauge, like in a VU meter, to show that the value is over range)

kmatch98:
This resolves: https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/issues/23

by adding a `limit_rotation` option to constrain needle rotation between min and max values.  Default is `True` to constrain rotation.

Also, I updated the docstrings to include the default values of most input parameters.

### DIFF
--- a/adafruit_displayio_layout/widgets/dial.py
+++ b/adafruit_displayio_layout/widgets/dial.py
@@ -48,47 +48,56 @@ class Dial(Widget):
 
     :param int width: requested width, in pixels
     :param int height: requested height, in pixels
-    :param int padding: keepout padding amount around the border, in pixels
+    :param int padding: keepout padding amount around the border, in pixels,
+     default is 12
 
-    :param float sweep_angle: dial rotation, in degrees, maximum value is 360 degrees
-    :param float start_angle: starting angle, in degrees.  Defaults
-     to `None` for symmetry along vertical axis.  Vertical is defined as 0 degrees.
+    :param float sweep_angle: dial rotation, in degrees, maximum value is 360 degrees,
+     default is 90 degrees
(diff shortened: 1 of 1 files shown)


jposada202020:
I review the changes mentioned in https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/pull/26/commits/5733d1d9be859e3de025ed8d5c995a6a991aacb5 on
```python
Adafruit CircuitPython 6.2.0-beta.4 on 2021-03-18; Seeeduino Wio Terminal with samd51p19
```

I did this changing the values in the simple_test_dial using ``-5`` and ``2000``. 
https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/blob/5733d1d9be859e3de025ed8d5c995a6a991aacb5/examples/displayio_layout_dial_simpletest.py#L24

Changes work as mentioned in this PR. Needle moves slower, but did not go under or over the dial.


#### Looks good to me

FoamyGuy (approved):
These changes look good to me. Thank you for testing @jposada202020.

Thank you for the new feature and documentation improvements @kmatch98

PR #29, opened 2021-03-26
Equalizer Widget

jposada202020:
Proposal for the Equalizer widget.

Suggestions are welcome. Just a simple widget that people can use to shows some sound spectrum

## Image
equalizer

## Initial test code
```python
import board
import displayio


def example():
    equa = Equalizer(
        x=100,
        y=100,
        width=100,
        height=100,
        number_bars=5,
        bar_width=10,
        background_color=0x22AA44,
        bar_line_color=0x2222DD,

    )
    my_group.append(equa)


display = board.DISPLAY
my_group = displayio.Group(max_size=3)

example()
display.show(my_group)
```
image


jposada202020:
Having problems with the fancyled import, I just import it following the examples in Adafruit. However,maybe I am doing something wrong. This is ready for review.

jposada202020:
@FoamyGuy @kmatch98 if you could take a look, after seeing some of the discussion in the flipselector PR and all the memory allocation problems I tried to make a widget simple but useful, we could build from here if people really use it and request features. Thanks

tannewt:
I think you should split out widgets into their own repos. They'll be easier to find that way. I'd expect this library to only have layout related classes.

jposada202020:
@tannewt No problem work on that this week

jposada202020:
Closing here, until new bundle is created

PR #31, opened 2021-04-04
Add Color picker Widget

jposada202020:
Allows the user to select a color using a wheel color widget

image


tannewt:
Separate repo for this too please.

jposada202020:
Ok I willl work on that this week :)

jposada202020:
Closing for now, until discussion regarding the graphics libraries move

kmatch98:
@jposada202020 I looked through your code but haven't run it yet.  One question, rather than do the fancy math to find the color and hue, can you instead query the specific pixel color in the bitmap?  Or is there some limitation of OnDiskBitmap that prevents you from getting the color value?

The reason I ask this: If you can pick the color directly from the bitmap, then you could always change bitmap to any file you want.

jposada202020:
@kmatch98, there is no particular reason, lack of knowledge on getting the color information with the OnDiskBitmap. There is always place for improvements 😃

kmatch98:
I actually have an open PR on being able to tweak the OnDiskBitmap palette, but that’s been a couple of weeks and I already forgot what’s possible or not. So that’s why I was curious if what you want to do is possible or not with OnDiskBitmap and whether there is something else you need. 

If you need a function in the core for OnDiskBitmap let me know and I may be able to help whenever I get back to this open PR: https://github.com/adafruit/circuitpython/pull/4438

jposada202020:
@kmatch98 thanks, Also another difficult part to do this, it was I have no function to put a color in a pixel, so I need to create a buffer and the save it to the SDCARD because I have not memory in the boards. The script to make the bitmaps works both in python and in CircuitPython. So that is another improvement that we could do. thanks for the info

PR #33, opened 2021-04-08
Move dependencies from autodoc mock imports to requirements.txt

makermelissa:
Since the JSON file looks in requirements, it's better to have them there.


FoamyGuy (approved):
Looks good to me, thank s for the fix. 

I Tested the sphinx doc building successfully.

Issue #34, opened 2021-04-21
Detailed documentation is missing from ReadTheDocs


kmatch98:
I noticed that the detailed documentation from the docstrings is no longer being generated into the ReadTheDocs. 

It's missing the graphics and all the long-form description of how these widgets are made.  

- One example is the `Widget` class definition that had a lot of details about how to use it.  
- Another example is the `Dial` widget where the graphics are missing.

**Q. Is it worth resolving this now or wait until these find their new home in the CircuitPython github pages?**

jposada202020:
Interesting I have the version of my repo for in readthedocs and is showing fine. 
https://circuitpython-displayio-jdm.readthedocs.io/en/latest/api.html

Maybe changes introduces after https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/commit/f5ab9c85549875db2cee020d1b6f730f252478af

I verify one of the libraries that I have updated last week an is showing fine with the nice text and verything.

jposada202020:
Could this be related
![image](https://user-images.githubusercontent.com/34255413/115720183-ce5c6080-a34a-11eb-9306-75bfb19914a3.png)

kmatch98:
I tried building the docs locally and they failed due to the missing "autodoc_mock_imports" in the `conf.py` file with error message:
```
Warning, treated as error:
autodoc: failed to import module 'grid_layout' from module 'adafruit_displayio_layout.layouts'; the following exception was raised:
No module named 'displayio'
```

Could this have been caused by: https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/commit/cd8728afae72d5bfa9e648ffb86ef9565f28ea25

Is there something that needs to be configured on my local machine to work like the docs building on the CI is working?

jposada202020:
As I suffered a lot with this, only the library requirements, But I have seen differences in the builts in the way the CI  and the locally. As proven in color picker library docs. 
I could bump my repo, and rebuild to see if the changes in https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/commit/cd8728afae72d5bfa9e648ffb86ef9565f28ea25 are the cause, what do you think @kmatch98 ? (just do not want to loose the beautiful documentation 😄 )

jposada202020:
@kmatch98 I Merged the changes and the nice documentation is still there and it works, so the changes are not the problem. Maybe something in the CI
[...]

PR #36, opened 2021-04-27
Solving parameters

jposada202020:
Solves #34 
@kmatch98 This solves the parameter Issue. Tested locally. Please verify


kmatch98 (approved):
The local build fails like this:
```
Warning, treated as error:
autodoc: failed to import module 'grid_layout' from module 'adafruit_displayio_layout.layouts'; the following exception was raised:
No module named 'displayio'
```

I suspect this is caused by the different impact of the `requirements.txt` versus the conf.py "mock-imports".  Let's merge this and let's see how it looks.

jposada202020:
Probably you will need to recreate the dev environment, as requirements change after the last release, or you could mock the import in you local directory and it will solve the problems. I verified the built and the documents are showing ok.

image

PR #38, opened 2021-04-30
docs_refactor_proposal

jposada202020:
This is the proposal for the new documentation refactor.

There are some things to change, I went to the first chapter more in deep adding the dynamic links, and other rst things. Did not spend a lot of time in the second chapter, no need in case we do not want to go this way.  Anyway you can comment, and see if it is a good idea or not. 

Please review and comment @FoamyGuy @kmatch98


kmatch98 (approved):
I think this is a great way to reorganize the documentation.  It separates the detailed documentation from the quick function and parameter details.  This update should make it easier to find what you're looking for without getting bogged down.

jposada202020:
Thanks, This is in accordance with the design guide, I will translate this PR to the CircuitPython Graphics sections once libraries are moved. Thanks @kmatch for reviewing :)

PR #39, opened 2021-05-24
Linted - didn't test

evaherrada:
@FoamyGuy want to test this since you're pretty familiar with this library?


jposada202020 (approved):
tested on
```python
Adafruit CircuitPython 6.2.0-beta.4 on 2021-03-18; Adafruit PyPortal Titano with samd51j20
```
works as expected

PR #40, opened 2021-07-01
Remove max_size parameter

lesamouraipourpre:
Remove usage of the `max_size` parameter now that it has been disabled in `shared-bindings`.

The examples have been tested on a Pynt:
```
Adafruit CircuitPython 6.3.0 on 2021-06-01; Adafruit PyPortal with samd51j20
```


FoamyGuy (approved):
looks good to me. Tested all examples on PyPortal with
```
Adafruit CircuitPython 7.0.0-alpha.4 on 2021-07-08; Adafruit PyPortal with samd51j20
```

Noticed an issue with IconWidget and IconAnimated, but I do not think it's related to this change specifically. Will work on trying to make a smaller reproduction of the issue containing only what is necessary to see it happen.

PR #41, opened 2021-07-10
fix OnDiskBitmap pixel_shader

FoamyGuy:
These changes are needed to support the new behavior of OnDiskBitmap from this PR: https://github.com/adafruit/circuitpython/pull/4823

Without these changes it causes IconAnimated and IconWidget to not be able to show the icons properly if they have on_disk=True parameter.

### DIFF
--- a/adafruit_displayio_layout/widgets/icon_widget.py
+++ b/adafruit_displayio_layout/widgets/icon_widget.py
@@ -59,7 +59,7 @@ class IconWidget(Widget, Control):
         if on_disk:
             self._file = open(self._icon, "rb")
             image = OnDiskBitmap(self._file)
-            tile_grid = TileGrid(image, pixel_shader=ColorConverter())
+            tile_grid = TileGrid(image, pixel_shader=getattr(image, 'pixel_shader', ColorConverter()))
         else:
             image, palette = adafruit_imageload.load(icon)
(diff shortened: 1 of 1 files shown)


adafruit_displayio_layout/widgets/icon_widget.py
```diff
@@ -59,7 +59,9 @@ def __init__(self, label_text, icon, on_disk=False, **kwargs):
         if on_disk:
             self._file = open(self._icon, "rb")
             image = OnDiskBitmap(self._file)
-            tile_grid = TileGrid(image, pixel_shader=ColorConverter())
+            tile_grid = TileGrid(
+                image, pixel_shader=getattr(image, "pixel_shader", ColorConverter())
+            )
         else:
```
lesamouraipourpre, line 68: Consider adding the following comment after the `tile_grid = ...` line
```python
            # TODO: Once CP6 is no longer supported replace the above line with below.
            # tile_grid = TileGrid(image, pixel_shader=image.pixel_shader)
```

lesamouraipourpre (approved):
Tested on `PyPortal` with `6.3.0` and `7.0.0-alpha.4` and it works correctly.

tannewt (approved):
Thank you!

PR #42, opened 2021-07-24
fix for content elements that dont allow setting width and height

FoamyGuy:
This fixes the issue noted here: https://github.com/adafruit/Adafruit_CircuitPython_MacroPad/issues/10

Label and BitmapLabel do not allow setting arbitrary width and height values. So we catch the exception and just use the size that the label already is. 

In the past Label and BitmapLabel had width and height class variables so those got set from GridLayout but did not actually change the visual appearance of the label. So catching this exception results in the exact behavior that we had previously.

### DIFF
--- a/adafruit_displayio_layout/layouts/grid_layout.py
+++ b/adafruit_displayio_layout/layouts/grid_layout.py
@@ -79,15 +79,20 @@ class GridLayout(displayio.Group):
                         ),
                     )
                 else:
-                    # try width and height properties.
-                    cell["content"].width = (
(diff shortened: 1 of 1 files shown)


lesamouraipourpre (approved):
LGTM

Tested on a PyPortal Pynt with `7.0.0-alpha.5`
Examples used:
DisplayIO_Layout/examples/displayio_layout_gridlayout_simpletest.py
MacroPad/examples/macropad_grid_layout.py (adapted to run on a PyPortal and simulate random key presses)

PR #46, opened 2021-08-06
adding get_cell function to GridLayout

### LINKED ISSUE #43: Provide a way to lookup elements by x,y grid position in GridLayout
Currently the sub-elements within a GridLayout can be accessed by index of the order that they were added:
```
my_grid_layout[0]
```
Will return the first element that was added to the grid so that you can access it's properties i.e.
```
my_grid_layout[0].color = 0xFFFFFF
```
would change the font color of a Label that was the first item added.

It would be great to provide an alternate way to lookup the elements in the grid based on their x,y cell location so that you could do something like:
```
my_grid_layout.get_element(0,0).color = 0xFFFFFF
```
and it would give you back the element in the first row / column without needing to worry about the order that the elements were added in.

FoamyGuy:
resolves #43 

Adding a new `get_cell()` function that allows you to lookup cell content objects by their coordinates within the GridLayout. Added a new example to illustrate this functionality.

### DIFF
--- a/adafruit_displayio_layout/layouts/grid_layout.py
+++ b/adafruit_displayio_layout/layouts/grid_layout.py
@@ -52,6 +52,7 @@ class GridLayout(displayio.Group):
         self.grid_size = grid_size
         self.cell_padding = cell_padding
         self._cell_content_list = []
+        self._cell_dictionary = {}
 
     def _layout_cells(self):
 
@@ -120,4 +121,24 @@ class GridLayout(displayio.Group):
             "cell_size": cell_size,
         }
         self._cell_content_list.append(sub_view_obj)
+        self._cell_dictionary[grid_position] = len(self._cell_content_list) - 1
         self._layout_cells()
+
+    def get_cell(self, cell_coordinates):
+        """
+        Return a cells content based on the cell_coordinates. Raises
+        KeyError if coordinates were not found in the GridLayout.
+
+        :param tuple cell_coordinates: the coordinates to lookup in the grid
+        :return: the displayio content object at those coordinates
+        """
+        if cell_coordinates in self._cell_dictionary:
+            return self._cell_content_list[self._cell_dictionary[cell_coordinates]][
+                "content"
+            ]
+
+        raise KeyError(
+            "GridLayout does not contain cell at coordinates {}".format(
+                cell_coordinates
+            )
+        )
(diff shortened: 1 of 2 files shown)


flavio-fernandes:
LGTM. Also may be nice to include link to the super awesome recording that shows all the steps performed.
https://youtu.be/ED0luF_q3ec

adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, x, y, width, height, grid_size, cell_padding):
         self.grid_size = grid_size
         self.cell_padding = cell_padding
         self._cell_content_list = []
+        self._cell_dictionary = {}
```
rsbohn: The `grid_position` is already stored in _cell_content_list. Maybe we can use that instead of adding a _cell_dictionary.

adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -120,4 +121,24 @@ def add_content(self, cell_content, grid_position, cell_size):
...
         self._layout_cells()
+
+    def get_cell(self, cell_coordinates):
+        """
+        Return a cells content based on the cell_coordinates. Raises
+        KeyError if coordinates were not found in the GridLayout.
+
+        :param tuple cell_coordinates: the coordinates to lookup in the grid
+        :return: the displayio content object at those coordinates
+        """
```
rsbohn, line 132: # create a list of grid_positions [(0,0), (0,1) ...]
    # index in cell_positions is the same as the index of the cell in self._cell_content_list.
    cell_positions = [cell['grid_position'] for cell in self._cell_content_list]
    if cell_coordinates in cell_positions:
        return self.cell_content_list[cell_positions.index(cell_coordinates)]

rsbohn (commented):
Suggest using grid position in  _cell_content_list instead of adding _cell_dictionary.

adafruit_displayio_layout/layouts/grid_layout.py
FoamyGuy: Ah yes we can look it up from the list. Good idea, I will change it to work that way.

rsbohn (approved):
I like it!

flavio-fernandes:
> I like it!

Me too. Thanks again for your time going over this code today @FoamyGuy !

Issue #45, opened 2021-08-01
Dial will need updated for new vectorio


FoamyGuy:
ref: [#5087](https://github.com/adafruit/circuitpython/pull/5087)

The `vectorio` API will be getting changed with regards to it's usage of palette. The Dial wiget that is currently in this library (although slated to move to it's own in circuitpython org) will need updated to reflect those changes in order to work properly.

https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/blob/cb263f13f09f3f42cb3183774cbf3bc1c46a4883/adafruit_displayio_layout/widgets/dial.py#L30

PR #47, opened 2021-08-21
remove annotation and dial widgets from this repo. 

FoamyGuy:
This removes Annotation and Dial widgets and associated examples and documentation from this repo. These now have their own repos over in the circuitpython organization and they've both been added to the CircuitPython_Org bundle.

All of the code and documentation removed with this PR is now available in these respective repos:
https://github.com/circuitpython/CircuitPython_Org_DisplayIO_Dial
https://github.com/circuitpython/CircuitPython_Org_DisplayIO_Annotation

The other widgets that are in this repo currently will get removed as soon as they are setup in new repos in circuitpython org and added to that bundle.

This PR also updates the pylintrc file to the latest version that cookie-cutter now generates.


FoamyGuy:
Forgot to mention in the original comment but I would consider #45 resolved once this is merged.

The new Dial repo has been updated already to support the latest version of the `vectorio` API

kattni (changes requested):
@FoamyGuy Are there any guides/examples outside the repo that need to be updated, or were these features never used elsewhere?

FoamyGuy:
@kattni to the best of my knowledge there are no guide or example usages of these two widgets outside of this repo. I grepped for both in the Learn guide repo and did not find any.

As far as I know the only one of the widgets that are currently in this repo that did get used for a guide / example is the IconWidget which was used in the Stream Deck project. This PR doesn't remove IconWidget, my intention there is to ensure that the project bundler used by Learn system will be able to find it properly in it's new home before removing it from here so that the transition will be seamless to folks who are following that guide.

kattni (approved):

kattni:
Thanks!

PR #53, opened 2021-11-24
Update grid_layout.py

gingershaped:
Added the ability to change the divider line color

### DIFF
--- a/adafruit_displayio_layout/layouts/grid_layout.py
+++ b/adafruit_displayio_layout/layouts/grid_layout.py
@@ -44,10 +44,12 @@ class GridLayout(displayio.Group):
       lines above. Row indexes are 0 based.
     :param Union[tuple, list] v_divider_line_cols: Column indexes to draw divider
       lines before. Column indexes are 0 based.
+    :param divider_line_color: The color of the divider lines (in hexadecimal)
 
     """
 
     # pylint: disable=too-many-arguments
+    # pylint: disable=too-many-instance-attributes
     def __init__(
         self,
         x,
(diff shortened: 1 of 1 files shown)


FoamyGuy:
@GingerIndustries thanks for adding this functionality. This is currently failing pylint due to `too-many-instance-attributes`. apparently we were previous right at the limit, and adding the new one pushed it over.

Some of the attributes could be collapsed into a dictionary or other object that can hold multiple things. But it seems to me that this would make the code more complex for the purpose of pleasing pylint. 

Personally I think it would be okay to disable the `too-many-instance-attributes` check for this class. But I'm open to input from others. If that does turn out to be the route we want to go with, do you know how to do that @GingerIndustries?

gingershaped:
It probably doesn't matter, as this will be compiled to an .mpy regardless. I could make the changes if you want.

FoamyGuy:
@GingerIndustries we will have to getting passing the CI checks one way or another before it can be merged. Even though the resulting library will get compiled to mpy we still run pylint and a few other checks on the .py version of the code using github actions.

Go ahead and add the pylint disable comment for that specific check. Let me know if you have any trouble, or need help with that.

gingershaped:
Okay, it's done.

FoamyGuy (approved):
Changes look good to me. Tested successfully on a PyPortal with a modified the gridlayout_dividers script in the examples directory. 

I tested hex colors i.e. `0xFF00FF` as well as tuple colors i.e `(255, 0, 255)` and both work as expected. I think internally displayio.Palette allows both to work in this case without any specific code needed in this library. 

Thanks for adding this new functionality @GingerIndustries

gingershaped:
Nice! Thanks,

PR #56, opened 2021-11-26
Handle missing bitmaptools

rsbohn:
If bitmaptools is missing an ImportError will be raised.
This handles the correct Error.


ladyada (approved):

ladyada:
didnt test!

FoamyGuy (approved):
Looks good to me. I confirmed on a TrinketM0 that the exception raised is an `ImportError` not a `NameError` So these changes do resolve the catching of the incorrect error type.

Thanks for the fix @rsbohn

rsbohn:
Thank you!

PR #57, opened 2021-11-26
allow missing vectorio

### LINKED ISSUE #51: vectorio import should be optional inside the widget module
Currently the import here:
https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/blob/f9438d2268248eff46ef4e85f686ac6e49f0f321/adafruit_displayio_layout/widgets/__init__.py#L10

makes `vectorio` a hard requirement for anything inside of the `widgets` module within this library. 

If I understand correctly the vecotorio import is only used if something calls `rectangle_helper()` which is also in this file. Wrapping this import with try/except would allow the module to be used in contexts where `vectorio` is not present i.e. using objects that extend `widget.Widget` with Blinka_Displayio (as long as those widgets don't need `rectangle_helper()` or anything else from `vectorio` specifically.

FoamyGuy:
resolves: #51 

These changes include:

- making vectorio requirement optional
- fix docstring of older example
- add new example that wouldn't be possible to run without this change.

### DIFF
--- a/examples/displayio_layout_pygame_display_switch_round.py
+++ b/examples/displayio_layout_pygame_display_switch_round.py
@@ -17,7 +17,7 @@ from adafruit_displayio_layout.widgets.switch_round import SwitchRound as Switch
 display = PyGameDisplay(width=320, height=240)
 
 # Make the display context
-main_group = displayio.Group(max_size=10)
+main_group = displayio.Group()
 display.show(main_group)
 
 switch_x = 30
--- a/examples/displayio_layout_pygame_display_switch_round.py
+++ b/examples/displayio_layout_pygame_display_switch_round.py
@@ -43,7 +43,7 @@ text_stroke = switch_stroke  # width of text lines
 touch_padding = 0  # Additional boundary around widget that will accept touch input
 
 animation_time = 0.2  # time for switch to display change (in seconds).
-                      # animation_time=0.15 is a good starting point
+# animation_time=0.15 is a good starting point
 display_text = True  # show the text (0/1)
 
 # initialize state variables
--- a/examples/displayio_layout_pygame_display_switch_round.py
+++ b/examples/displayio_layout_pygame_display_switch_round.py
@@ -1,0 +1,83 @@
+# SPDX-FileCopyrightText: 2021 Tim C
+#
+# SPDX-License-Identifier: MIT
+"""
+Make a GridLayout with some Labels in it's cells.
+Displayed with Blinka_Displayio_PyGameDisplay
+
+Requires: https://github.com/FoamyGuy/Blinka_Displayio_PyGameDisplay
+"""
+import displayio
+import pygame
+from blinka_displayio_pygamedisplay import PyGameDisplay
+from adafruit_displayio_layout.widgets.switch_round import SwitchRound as Switch
+
+
+# Make the display context. Change size if you want
+display = PyGameDisplay(width=320, height=240)
+
+# Make the display context
+main_group = displayio.Group(max_size=10)
+display.show(main_group)
+
+switch_x = 30
+switch_y = 30
+switch_radius = 20
+
+switch_fill_color_off = (200, 44, 200)
+switch_fill_color_on = (0, 100, 0)
+
+switch_outline_color_off = (30, 30, 30)
+switch_outline_color_on = (0, 60, 0)
+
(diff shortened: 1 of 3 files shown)


examples/displayio_layout_pygame_display_switch_round.py
```diff
@@ -0,0 +1,83 @@
...
+import pygame
+from blinka_displayio_pygamedisplay import PyGameDisplay
+from adafruit_displayio_layout.widgets.switch_round import SwitchRound as Switch
+
+
+# Make the display context. Change size if you want
+display = PyGameDisplay(width=320, height=240)
+
+# Make the display context
+main_group = displayio.Group(max_size=10)
```
rsbohn: max_size is deprecated

rsbohn:
When I tried running with Blinka it complained that I was trying to put an invalid color into a palette.

switch_round.py line 724 self._switch_circle.fill was {'transparent': False, 'rgb888': 16777215, 'rgba': (255, 255, 255, 255)}. If I used ..fill['rgba'] it would run.

examples/displayio_layout_pygame_display_switch_round.py
FoamyGuy: good catch, thank you. I copied this example from an older starting script. This is removed with the latest commit.

FoamyGuy:
> When I tried running with Blinka it complained that I was trying to put an invalid color into a palette.

@rsbohn Strange. I am seeing that error now to. I am fairly positive that I ran this example successfully before making the PR. I must have had a different version of something at that time that allowed it to work. 

I'm not sure what you mean when you mentioned "If I used ..fill['rgba'] it would run". Is this a change you made inside of the example? or in the switch_round class or somewhere else? Can you post the code that you changed to when you got it to run?

FoamyGuy:
Interestingly I was able to test both `displayio_layout_switch_multiple.py` and `displayio_layout_switch_simpletest.py` on a PyPortal. So whatever the root of this issue is, it's either only an issue in the Blinka_DisplayIO context, or else caused by something in the new example that isn't present in the others.

FoamyGuy:
@rsbohn I figured out what you mean I think. I see that Palettes in Blinka_DisplayIO are containing dictionaries for each item, whereas in the core the palette items are literal ints. So I can see how using `fill['rgba']` inside of SwitchRound class would allow the code to run in Blinka_DisplayIO, but I think it would crash if/when it ran on a microcontroller with core displayio.

I think the fix will be to try to change Blinka_DisplayIO to make it return the same value as core displayio when using square brackets to __get_item__ out of a Palette.

I am still a bit perplexed as to when this difference started occurring. I'm pretty sure I've had this code execute successfully before, but I don't think it could with the Palette behavior is like this. I took a peek at recent PRs in Blinka_DisplayIO but don't see anything that seems like it would have been relevant to this difference so I'm still unsure what could have changed in my local environment to cause this to start happening.

FoamyGuy:
I did find a fix inside of Blinka_DisplayIO for this. I changed `__getitem__` inside of Palette to this:
```
    def __getitem__(self, index):
        if not 0 <= index < len(self._colors):
            raise ValueError("Palette index out of range")
        return self._colors[index]["rgb888"]
```
and it allows this example to run successfully (and equalizes the behavior between core displayio and Blinka_DisplayIO). I think that is the best solution for this issue, but I would like to wait for input from Melissa to see if there was a specific reason for the difference, or if changing it as above could cause other issues.

examples/displayio_layout_pygame_display_switch_round.py
```diff
@@ -0,0 +1,83 @@
+# SPDX-FileCopyrightText: 2021 Tim C
+#
+# SPDX-License-Identifier: MIT
+"""
+Make a GridLayout with some Labels in it's cells.
```
rsbohn, line 5: Update the docstring to match the code (SwitchRound demo for PyGameDisplay).

rsbohn:
Sorry I didn't respond sooner. This should be okay to merge, but I won't be able to test it until #[Adafruit_Blinka_Displayio #73](https://github.com/adafruit/Adafruit_Blinka_Displayio/pull/73) is resolved.
It looks like the one function in widgets/__init__.py is only used in the cartesian.py widget, but there may be other uses outside of this module.

PR #58, opened 2021-11-28
Cell anchor point

FoamyGuy:
This change adds a new feature `cell_anchor_point` which can be specified by an argument in the initialization of the GridLayout, and overridden on a per cell basis by passing an argument with the same name to the `add_content()` function.

`cell_anchor_point` will cause the content of the cell to be anchored to a spot relative to the float values of the anchor_point. e.g:

`(0.0, 0.0)` which is default will anchor content to the top left within each cell. This is the same behavior that GridLayout has today without these changes.

`(1.0, 1.0)` would anchor content to the bottom right within each cell

`(0.5, 0.5)` would anchor content into the center of each cell

`(0.5, 0.0)` would anchor content into the top center of each cell.

Here is a screenshot showing the usage of this new feature to center content within each cell:
image


kattni (approved):
Did not test. Looks good!

PR #62, opened 2021-12-16
Widget cartesian  update line  bugfix

### LINKED ISSUE #61: `Cartesian` how to use the `update_line` function?
there is a `update_line` function in the `Cartesian` class.
but i have not found any documentation about it -

i would expect to be able to draw some kind of linegraph with multiple points.
but [looking at the code](https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/blob/main/adafruit_displayio_layout/widgets/cartesian.py#L489) and testing that does not happen..

it seems that the first time the function is called it setups `self.plot_line_point` and adds the given point.
and after this it draws a line from this first point to the given point.
resulting in a *star-like* graph.

i tested and the only thing missing is a `self.plot_line_point.append((local_x, local_y))` after the [call to `bitmaptools.draw_line(`](https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/blob/main/adafruit_displayio_layout/widgets/cartesian.py#L508)

my minimal testcode:
```
# SPDX-FileCopyrightText: 2021 Stefan Krüger
#
# SPDX-License-Identifier: MIT
#############################
"""
Simple Test for line-ploting with the Cartesian widget.
"""

import time
import board
import displayio
from adafruit_displayio_layout.widgets.cartesian import Cartesian

# create the display on the PyPortal or Clue or PyBadge (for example)
display = board.DISPLAY


# pybadge display:  160x128
# Create a Cartesian widget
# https://circuitpython.readthedocs.io/projects/displayio-layout/en/latest/api.html#module-adafruit_displayio_layout.widgets.cartesian
my_plane = Cartesian(
    x=30,  # x position for the plane
    y=2,  # y plane position
    width=128,  # display width
    height=105,  # display height
    xrange=(0, 90),
    yrange=(0, 90),
)

my_group = displayio.Group()
my_group.append(my_plane)
display.show(my_group)  # add high level Group to the display

positions = [
    (0, 0),
    (10, 10),
    (20, 50),
    (30, 60),
    (40, 40),
    (50, 80),
    (60, 20),
    (70, 60),
    (80, 30),
    (90, 90),
]

print("displayIO__layout_widget_cartesian_lineplot.py")
for x, y in positions:
    my_plane.update_line(x, y)
    time.sleep(1)

while True:
    pass
```

@FoamyGuy  @kattni should i submit a pull request for this?

s-light:
fixes #61
and fixes misalignment of graph to axis.
there was a factor missing..


adafruit_displayio_layout/widgets/cartesian.py
```diff
@@ -457,54 +457,162 @@ def _draw_pointers(self, x: int, y: int) -> None:
...
+                self._yrange[1],
+            )
         )
-
-        if local_x >= 0 or local_y <= 100:
-            if self._update_line:
-                self._draw_pointers(local_x, local_y)
-                self._update_line = False
+        local_x, local_y = self._calc_local_xy(x, y)
+        print(
```
FoamyGuy: same with this print()

adafruit_displayio_layout/widgets/cartesian.py
```diff
@@ -457,54 +457,162 @@ def _draw_pointers(self, x: int, y: int) -> None:
...
         :param int x: ``x`` coordinate in the local plane
         :param int y: ``y`` coordinate in the local plane
         :return: None
         rtype: None
         """
-        local_x = int((x - self._xrange[0]) * self._factorx) + self._nudge_x
-        local_y = (
-            int((self._yrange[0] - y) * self._factory) + self.height + self._nudge_y
+        print("")
+        print(
```
FoamyGuy: lets remove the print statements, or add a `verbose` argument used to enable them.

FoamyGuy (commented):
Thanks for working on this @s-light

adafruit_displayio_layout/widgets/cartesian.py
s-light: yeah - sorry ;-)
i added an verbose flag

examples/displayio_layout_cartesian_dev.py
```diff
@@ -0,0 +1,69 @@
+# SPDX-FileCopyrightText: 2021 Stefan Krüger
+#
+# SPDX-License-Identifier: MIT
+#############################
+"""
+This is a basic demonstration of a Cartesian widget for line-ploting
```
FoamyGuy: Do we need to keep both `displayio_layout_cartesian_dev.py` and `displayio_layout_cartesian_lineplot.py` ? They seem to be doing mostly the same thing. I think I would lean towards just keeping `displayio_layout_cartesian_lineplot.py` and if you want you can add the pointer behavior and different colors to it (though I think it's also fine to leave those out, either way is okay).

adafruit_displayio_layout/widgets/cartesian.py
```diff
@@ -465,46 +593,31 @@ def update_pointer(self, x: int, y: int) -> None:
...
+        if not self._pointer:
+            self._draw_pointers(
+                self.plot_line_point[-1][0],
+                self.plot_line_point[-1][1],
+            )
+        else:
+            self._pointer.x = self.plot_line_point[-1][0]
+            self._pointer.y = self.plot_line_point[-1][1]
 
     def update_line(self, x: int, y: int) -> None:
```
FoamyGuy: I think we should rename this to `draw_line()` or `add_line()` instead of `update_line()` It looks like it was a previously existing function but maybe didn't function as expected in the past?

With the new behavior it now adds a new line to the graph each time you call it. Rather than moving the previously created line.

The `update_pointer()` function changes the location of the pointer (i.e. removes it from where it used to be and adds it in the location so that it changes location)

So with the current name `update_line()` there seems to be different behavior between `update_line()` and `update_pointer()` even though they have the same prefix in their name. I think we should either match the behavior, or choose a different name for `update_line()`. I would lean towards the latter because I think it's cool to be able to add multiple lines to the chart. Maybe at some point we could even make possible to add multiple points to the chart as well with a `draw_point()` or `add_point()` function.

FoamyGuy (changes requested):
Thanks for working on this @s-light. I have a few small requests but this is looking good to me for the most part.

Tested successfully on PyPortal 7.1.0-beta.3

examples/displayio_layout_cartesian_dev.py
s-light: yeah - thati will remove the *_dev.py.
and keep the *simple / minimal* plot.

adafruit_displayio_layout/widgets/cartesian.py
s-light: yeah - i did not touch the *public api*...
and you are right - it did not really work - at least it definitly also did not update the line... it draw a new one each time...
so a api name change at this point makes sens to me!
i like the `add_line()` version - as it most accurately describes what happens internally...
i also added a `clear_lines()` function. that just does what it says ;-)
and so you have the option to also do some sort of update (draw / clear / draw / ...)

examples/displayio_layout_cartesian_lineplot.py
```diff
@@ -0,0 +1,63 @@
...
+    (8, 8),
+    (9, 9),
+    (10, 9),
+    (10, 10),
+]
+
+print("examples/displayio_layout_cartesian_lineplot.py")
+
+# first point without a wait.
+my_plane.update_line(0, 0)
```
FoamyGuy: @s-light Thank you for renaming it. There is one spot here that needs updated to the new name `add_line()`.

This looks good to me after that. I retested latest version(with this change) on pyportal.

s-light:
@FoamyGuy i just questioned myself if it makes more sens to change the nameing to `line_add()` and `lines_clear()` or `plot_add_line` and `plot_clear` or `graph_add_line` and `graph_clear`
this way it feels a little bit better *ordered*....
what you are thinking?

FoamyGuy:
@s-light I do like the idea of adding "plot" to the names to make it distinct from the axis lines. How about `add_plot_line()` and `clear_plot_lines()`?

s-light:
so you prefer the *action first* naming?
(i personally prefer *object subobject action* naming... this also help for simple autocomplete - all the function/attribut names start with the object name the belong to...)
i am fine with this. :-)

FoamyGuy (approved):
LGTM

@s-light I do prefer action first naming. However I do see the upside you mention about autocomplete. I may be spoiled a bit in that regards. I think PyCharm autocomplete is good about showing options even if I enter them out of order so if I start typing something like `my_plane.line` I will still get `add_plot_line` and `clear_plot_lines` in the list of possible completions even though they don't begin with "line" like I started typing.

PR #65, opened 2021-12-16
Missing int conversions on cell_anchor_point math

Neradoc:
In multiple places where `cell["cell_anchor_point"] * X` is used, the int conversion is missing, causing an error when an anchor_point is set to `(0.5, 0.5)` for example. I only encountered the issue with lines 179-182, but I think the others are required too.
```
  File "code.py", line 9, in <module>
  File "code_pyportal.py", line 285, in switch_page
  File "/lib/adafruit_displayio_layout/layouts/grid_layout.py", line 413, in add_content
  File "/lib/adafruit_displayio_layout/layouts/grid_layout.py", line 182, in _layout_cells
  File "/lib/adafruit_displayio_layout/widgets/widget.py", line 297, in anchored_position
  File "/lib/adafruit_displayio_layout/widgets/widget.py", line 234, in _update_position
TypeError: can't convert float to int
```
I didn't look into the math but I wonder, should it use `round()` rather that `int()` ? (which always rounds down)

### DIFF
--- a/adafruit_displayio_layout/layouts/grid_layout.py
+++ b/adafruit_displayio_layout/layouts/grid_layout.py
@@ -176,10 +176,10 @@ class GridLayout(displayio.Group):
                     cell["content"].anchored_position = (
                         int(grid_position_x * self._width / grid_size_x)
                         + self.cell_padding
-                        + (cell["cell_anchor_point"][0] * _measured_width),
+                        + int(cell["cell_anchor_point"][0] * _measured_width),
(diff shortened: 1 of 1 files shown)


FoamyGuy (approved):
looks good to me. Thanks @Neradoc.

Tested on pyportal

FoamyGuy:
I think it should always round down because we want to avoid if possible placing things outside of the display if they round up to a number higher than display size.

PR #67, opened 2021-12-16
handle float locations

### LINKED ISSUE #66: widget.py - TypeError: can't convert float to int
Re this thread:
https://forums.adafruit.com/viewtopic.php?f=47&t=186239

Added a dump of the values just before this line:
https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/blob/48b512a391732c03aff8556a0a2d688777e672a9/adafruit_displayio_layout/widgets/widget.py#L230
thusly:
```python
print(self._anchored_position[0], self._anchor_point[0], self._bounding_box[2], self._bounding_box[0])
```
and get:
```python
6.0 0.0 93 0
```

So some of the values are float making the result also float. `displayio.Group.x` requires `int`.

FoamyGuy:
resolves: #66 

Tested on pyportal and Feather RP2040 with 3.5TFT

### DIFF
--- a/adafruit_displayio_layout/widgets/widget.py
+++ b/adafruit_displayio_layout/widgets/widget.py
@@ -228,12 +228,12 @@ class Widget(displayio.Group):
         """
 
         if (self._anchor_point is not None) and (self._anchored_position is not None):
-            self.x = (
+            self.x = int(
                 self._anchored_position[0]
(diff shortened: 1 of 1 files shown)


FoamyGuy:
@caternuson want to try this one out when you can?

caternuson:
That looks like the same change I did locally when writing the issue as a quick check, and it worked.

PR #68, opened 2022-01-03
transparency in icon image. Background color for label.

FoamyGuy:
This adds two new features to the IconWidget both exposed as arguments to the constructor

- `label_background` which can be set to a hex color i.e. `0x888888` and will get set as the color in the  background of the icon label.
- `transparency_index` which can be set to an integer index within the palette of the icon bmp image. If set this will call `make_transparent()` on the given index to make it not get drawn on the display.

Both of these new features are used in the PyPortal Busy Simulator app that I am working on a guide for.


FoamyGuy:
Initially this failed CI due to sphinx not being able to find `BDF` class. In the latest commit I added bitmap_font library to requirements.txt in order to resolve that and make it passing CI.

askpatrickw (approved):
I gave this a quick review. Looks good.
I do see the docstring nicely covers all the new parameters as well.

Not tested.

PR #69, opened 2022-03-26
adding PageLayout and examples

FoamyGuy:
A new type of layout that allows you to organize your UI into pages that you can change between using name, index, or prev/next functions.

### DIFF
--- a/adafruit_displayio_layout/layouts/page_layout.py
+++ b/adafruit_displayio_layout/layouts/page_layout.py
@@ -124,10 +124,10 @@ class PageLayout(displayio.Group):
 
         self._check_args(page_name, page_index)
 
-        if page_index:
+        if page_index is not None:
             return self._page_content_list[page_index]
 
-        if page_name:
+        if page_name is not None:
             for cell in self._page_content_list:
                 if cell["page_name"] == page_name:
                     return cell
--- a/adafruit_displayio_layout/layouts/page_layout.py
+++ b/adafruit_displayio_layout/layouts/page_layout.py
@@ -188,6 +188,14 @@ class PageLayout(displayio.Group):
     def showing_page_name(self, new_name):
         self.show_page(page_name=new_name)
 
+    @property
+    def showing_page_content(self):
+        """
+        The content object for the currently showing page
+        :return Displayable: showing_page_content
+        """
+        return self._page_content_list[self._cur_showing_index]["content"][0]
+
     def next_page(self, loop=True):
         """
         Hide the current page and show the next one in the list by index
--- a/adafruit_displayio_layout/layouts/page_layout.py
+++ b/adafruit_displayio_layout/layouts/page_layout.py
@@ -1,0 +1,218 @@
+# SPDX-FileCopyrightText: 2022 Tim Cocks
+#
+# SPDX-License-Identifier: MIT
+
+"""
+`page_layout`
+================================================================================
+
+A layout that organizes pages which can be viewed one at a time.
+
+
+* Author(s): Tim Cocks
+
+Implementation Notes
+--------------------
+
+**Hardware:**
+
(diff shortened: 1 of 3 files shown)


adafruit_displayio_layout/layouts/page_layout.py
```diff
@@ -0,0 +1,218 @@
...
+        KeyError if the page was not found in the PageLayout.
+
+        :param string page_name: the name of the page to lookup
+        :param int page_index: the index of the page to lookup
+        :return: the displayio content object at those coordinates
+        """
+
+        self._check_args(page_name, page_index)
+
+        if page_index:
```
rsbohn: This should be `if page_index is not None:`

adafruit_displayio_layout/layouts/page_layout.py
```diff
@@ -0,0 +1,218 @@
...
+            return self._page_content_list[page_index]
+
+        if page_name:
```
rsbohn: This should be `if page_name is not None:`

rsbohn (changes requested):

rsbohn:
Consider adding a property to retrieve the current page content:

    showing_page_content -> Displayable

Make it easy for the user to retrieve the object they added using `add_content()` without any wrappers.

The user shouldn't need to know about `_page_group` and `sub_view_obj`, these are implementation details.

adafruit_displayio_layout/layouts/page_layout.py
FoamyGuy: Good catch thank you. Fixed with latest commit

adafruit_displayio_layout/layouts/page_layout.py
FoamyGuy: Good catch thank you. Fixed with latest commit

FoamyGuy:
> Consider adding a property to retrieve the current page content:
> 
> ```
> showing_page_content -> Displayable
> ```
> 
> Make it easy for the user to retrieve the object they added using `add_content()` without any wrappers.
> 
> The user shouldn't need to know about `_page_group` and `sub_view_obj`, these are implementation details.

Good idea, I like that functionality.  Thanks for trying this out and suggesting improvements. I'll add another commit with this new property

FoamyGuy:
Also I like the idea of a `Displayable` typing protocol (I think that is the right term). Lots of functions can accept this type of argument. Originally TileGrid and Group were the only that existed but now the vectorio shapes exist and are similar. Having this type defined somewhere will allow us to more easily update it if we end up adding any more types that can be shown directly like this.

FoamyGuy:
Latest commit adds the `showing_page_content` property as well as an illustration of it's usage in the advanced example script

FoamyGuy:
Not sure of the cause of CI failure. It passes locally for me:
image

Will look into this a bit later on today.

FoamyGuy:
I think this is the root caused of the CI failure. https://github.com/psf/black/issues/2964 Looks like there are a few options to get past it but it might be something we'll need to decided on across all of the libraries if I am understanding correctly.

FoamyGuy:
Indeed the Black / Click version issue is what was causing CI to fail. There was a commit yesterday to change the version used and I just re-ran the jobs and now it's passing :tada:

aioue:
**LGTM.**

Tested code on a Titano successfully.

Test process:
- Loaded `page_layout.py` onto a Titano
- Ran the 2 test files using adapted import line `from page_layout import PageLayout` (hobbyist python user)

Test results:
- `simpletest` PASS. Rendered both pages successfully and switches between them.
- `advancedtest` PASS. Rendered all three pages successfully and modifies the second page as expected.

Notes:
- Feels faster to load the code and render the text and shapes than other Learning System templates 🚀  Nice work FoamyGuy 👏

PR #72, opened 2022-05-14
align init() error messages

rsbohn:
Match error messages to the required init() parameters


FoamyGuy (approved):
Looks good, thank you for the fix.

PR #73, opened 2022-05-15
Hotplug sensor examples

PaulskPt:
This is my first PR. I was guided all the way towards this PR by  @FoamyGuy. 
This PR adds two examples that both have a functionality to hotplug external sensors. 
Both examples use the new TabLayout class that @FoamyGuy recently has added to the Adafruit_CircuitPython_DisplayIO_Layout library. The second example uses the Touch capabilities of the PyPortal.

The particularities of the in this PR presented 'hotplug sensor' algorithm are:
- when a sensor is disconnected the script will not crash. The script will notify the user: a) that the sensor has found disconnected; ask the user to check the wiring;
- In the mean time the main loop of the script in main() will continuously check for reconnection;
- If the sensor is reconnected the script will continue to get data from the sensor as the script did before.
The first example uses two external devices/sensors:
a) an Adafruit TMP117 sensor;
b) an external RTC module, DS3231.
both connected through I2C.

The second example is an expanded version of the first example.
As the first example also this script is able to read data from an external temperature sensor.
This script is able to use an external RTC.  The preferred setting is to use a NTP server.
The script uses the Adafruit TouchScreen library. For the touch of the Tabs to be reliable a calibration is advised. I made use of the calibration script by CedarGrove Studios ( https://github.com/CedarGroveStudios/Touchscreen_Calibrator/blob/main/touch_calibrator_built_in.py).
The calibration results have to be inserted at line 410 ( ts = adafruit_touchscreen.Touchscreen()), replacing what is there at 'calibration=((6815, 60095), (10520, 58007)),'
Both example scripts use a gVars class to set/get globally used variables. This eliminates the need to use global variables.
The gVars class has been created because we were confronted with pylint warnings about using global variables.
Both example scripts show the date and time in the Tab 'Three'. 
Both examples have a gVars flag "temp_in_fahrenheit". When set to 'True' the temperature will be displayed in degrees Fahrenheit. When set to 'False' in degrees Celsius. See image #4 below.
They display the temperature from the sensor in Tab 'Four'.
The second example script will notify the user in Tab 'Three' how many mm:ss a next NTP synchronization is due. Current resync is every 5 minutes.
This is done because we experienced that the internal clock runs ahead in time. Resync with NTP is crucial.
The second example, because of connecting to the internet, needs user-filled-in data from the file secrets.py.
A file secrets.py with default, unpersonalized, values is added to this PR.
This example also uses as a background image the image: BGimage4.bmp which is inside the subfolder /images.
About debugging: Both example scripts use a gVars variable "my_debug" which, when set to 'True' will have the script output more info about it's state/situation (I prefer to call it: 'debug information'). In the first example this flag is set in line 146. In the second example in line 172.
Initial clock setting: in both examples the clock will be set at startup by setting the globally used gVars variable "default_dt". In the first example this is in line 149. In the second example in line 175. The second example, when not using an external RTC device (determined by the setting of gVars flag "use_ntp" in line 200), it will use the date and time from an NTP server. The internal clock is set in line 341: "ntp.set_time(tz_offset)". If exists a valid time from NTP then  "default_dt" is set in line 363.
About performance: The time for the second example script to become in a 'Ready...' state it takes about 54 seconds from powering up. Tapping on Tab 'Four' to display the temperature value - only the first time - takes about 5 seconds.  After this one can tap from one Tab to the other with no significant delays.
ToDo: investigate/experiment how to improve the time from boot to the 'Ready...' moment;
During a recent Deep Dive with @FoamyGuy, viewers of this stream were enthousiastic about this sensor 'hotplug' feature. That is why @FoamyGuy advised me to create this PR.

The images below are taken while running the second example (the one using date and time from an NTP server).

IMG_5112

IMG_5113

IMG_5114

2022-05-16_temp_in_degs_F


examples/hotplug_sensor_examples/displayio_layout_tab_layout_hotplug_ext_rtc_temp_sensor_test.py
```diff
@@ -0,0 +1,656 @@
...
+display = board.DISPLAY
+# display.rotation = 90
+display.rotation = 0
+
+# create and show main_group
+main_group = displayio.Group()
+display.show(main_group)
+
+# fon.gvars bitmap_font.load_font("fonts/Helvetica-Bold-16.bdf")
+font_arial = bitmap_font.load_font("/fonts/Arial-16.bdf")
```
FoamyGuy, line 229: Lets make the custom font optional so that the example will still work if the user doesn't have the required font file. This way they can still see how to use a custom font. But if they don't have it, can still run the example without changing the code. 

I tested successfully changing to the `font_arial` setup to be like this:
```py
try:
    font_arial = bitmap_font.load_font("/fonts/Arial-16.bdf")
except OSError as e:
    # font not found. Fall back on built-in
    font_arial = terminalio.FONT
```

We can also get rid of the commented out line with the helvetica font setup.

examples/hotplug_sensor_examples/displayio_layout_tablayout_hotplug_temp_sensor_datetime_fm_NTP_touch_and_gVars_test.py
```diff
@@ -0,0 +1,996 @@
...
+"""
+# -----------------------------------
+
+# create and show main_group
+main_group = displayio.Group()  # The Main Display Group
+
+display.show(main_group)
+
+# font = bitmap_font.load_font("fonts/Helvetica-Bold-16.bdf")
+font_arial = bitmap_font.load_font("/fonts/Arial-16.bdf")
```
FoamyGuy, line 460: I think we should either make the arial font usage optional by have the try/except block here and falling back to built-in font. OR if we want to keep the arial font and make it non-optional we need to add a copy of this font file to the repo, but we'll also need the appropriate license for it for the reuse checks to pass.

I lean toward just making it optional, but am open to either solution.

examples/hotplug_sensor_examples/displayio_layout_tablayout_hotplug_temp_sensor_datetime_fm_NTP_touch_and_gVars_test.py
```diff
@@ -0,0 +1,996 @@
...
+# add it to the group that is showing on the display
+main_group.append(test_page_layout)
+# test_page_layout.tab_tilegrids_group[3].x += 50
+# ---------- Text Boxes ------------- #
+# Set the font and preload letters
+# font = bitmap_font.load_font("/fonts/Arial-16.bdf")  # was: Helvetica-Bold-16.bdf")
+# font.load_glyphs(b"abcdefghjiklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890- ()")
+glyphs = b' "(),-.0123456789:ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
+font_arial.load_glyphs(glyphs)
+font_arial.load_glyphs(("°",))  # a non-ascii character we need
```
FoamyGuy, line 651: If we go the route of making the arial font usage optional then we need to either remove the load_glyphs calls here or wrap them in an if statement because the built-in font doesn't have this function.

I successfully tested the following logic for it:
```py
if hasattr(font_arial, "load_glyphs"):
    glyphs = b' "(),-.0123456789:ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
    font_arial.load_glyphs(glyphs)
    font_arial.load_glyphs(("°",))  # a non-ascii character we need
``

FoamyGuy (changes requested):
Small change request to make the usage of Arial font optional or if it remains non-optional then we'll need to add the font file and a license for it to the repo.

Beyond that, these changes are looking good to me. Tested successfully on PyPortal Titano.

examples/hotplug_sensor_examples/displayio_layout_tablayout_hotplug_temp_sensor_datetime_fm_NTP_touch_and_gVars_test.py
FoamyGuy: The Arial font BDF file that I found in another project used in a learn guide project did actually already have license info included in it. So that goes a long way toward changing my mind to leaning toward including the font file and it's license in the examples directory rather than making it optional.

Here is the one I found and tested with: https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/main/PyPortal_OpenWeather/fonts/Arial-16.bdf

FoamyGuy (approved):
Looks good to me. Tested successfully on PyPortal Titano. @PaulskPt Thanks again for making this example and working through all of the checks and things to share it in repo.

Issue #75, opened 2022-10-02
Zero animation time for `SwitchRound` causes `ZeroDivisionError`


The-Compiler:
The docs [claim](https://docs.circuitpython.org/projects/displayio-layout/en/latest/api.html#feature-summary) that it's possible to set `animation_time` to zero:

> **animation:** Set `animation_time` to the duration (in seconds) it will take to transition the switch, set zero if you want it to snap into position immediately

However, the code doesn't actually handle that case, thus leading to a `ZeroDivisionError` here:

https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/blob/aa66d0fdd6b02e91b8e138dbfc12a7b3d04d3af3/adafruit_displayio_layout/widgets/switch_round.py#L767

Traceback:

```pytb
Traceback (most recent call last):
  [...]
  File "adafruit_displayio_layout/widgets/switch_round.py", line 830, in value
  File "adafruit_displayio_layout/widgets/switch_round.py", line 792, in selected
  File "adafruit_displayio_layout/widgets/switch_round.py", line 767, in _animate_switch
```

tekktrik:
Adding this a good first issue so maybe someone will pick it up for Hacktoberfest.  The fix is to make sure that if the animation time is zero that the animation process is skipped.  This should look something like a check to see if `_animation_time` is zero, and if it is, calculate the final position, and avoid dividing by the animation time entirely.

I'll pick this up soon-ish if no one does soon!

The-Debarghya:
Can you specify which files are involved?
I can work on this then.

The-Compiler:
@The-Debarghya See the file I linked in my issue description - that `_animate_switch` method should be the only place needing a change. Often, changes will also require corresponding doc or tests changes. However, in this case, the docs are already correct, and there seem to be no existing tests (so I'm assuming you won't be expected to add a new one for this change).

The-Debarghya:
Ok thanks, I'm trying to fix this.

tekktrik:
Fixed via #76

PR #76, opened 2022-10-07
handle ZeroDivisionError

The-Debarghya:
As mentioned in #75 if animation_time is set to 0, no animation should occur.

### DIFF
--- a/adafruit_displayio_layout/widgets/switch_round.py
+++ b/adafruit_displayio_layout/widgets/switch_round.py
@@ -772,10 +772,11 @@ class SwitchRound(Widget, Control):
                         1 - (elapsed_time) / self._animation_time
                     )  # fraction from 0 to 1
                 else:
-                    position = (elapsed_time) / self._animation_time  # fraction from 0 to 1
+                    # fraction from 0 to 1
+                    position = (elapsed_time) / self._animation_time  
 
-                # Update the moving elements based on the current position
-                # apply the "easing" function to the requested position to adjust motion
+            # Update the moving elements based on the current position
+            # apply the "easing" function to the requested position to adjust motion
                 self._draw_position(easing(position))  # update the switch position
 
             # update the switch value once the motion is complete
--- a/adafruit_displayio_layout/widgets/switch_round.py
+++ b/adafruit_displayio_layout/widgets/switch_round.py
@@ -754,6 +754,9 @@ class SwitchRound(Widget, Control):
             # Determines the direction of movement, depending upon if the
             # switch is going from on->off or off->on
 
+            if self._animation_time == 0:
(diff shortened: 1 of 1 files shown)


tekktrik:
Hi, I'll take a look at this but I think it may need a small tweak to get it to work.

FoamyGuy (changes requested):
Thanks for working on this @The-Debarghya!

The current way it's coded only handles when the switch is turning off (value going to False). It won't ever turn the switch back on this way because it's only calling draw_position(0) and never draw_position(1).

I tested this out on a Feather ESP32-S2 TFT and confirmed that is the current behavior.

I think it needs an `if` to check if current value is on or off then it can determine which position to draw. It also needs to allow the last part of the function execute. The part that is setting self._value to it's new state. So we need this not to `break` but instead to use `else` for the logic when animation_time > 0 so that it will run in that case, and we'll still get down to the _value updates if animation_time is zero.

I made those tweaks and it does seem to work as intended after that. This is the code that I ended up with:
```py
# if it's instant instead of animating
if self._animation_time == 0:
    print(f"0 anim time: {self._value}")
    if not self._value:
        position = 1
        self._draw_position(1)
    else:
        position = 0
        self._draw_position(0)

else:  # we need to animate over time
    # constrain the elapsed time
    elapsed_time = time.monotonic() - start_time
    if elapsed_time > self._animation_time:
        elapsed_time = self._animation_time

    # Determines the direction of movement, depending upon if the
    # switch is going from on->off or off->on
    if self._value:
        position = (
            1 - (elapsed_time) / self._animation_time
        )  # fraction from 0 to 1
    else:
        position = (elapsed_time) / self._animation_time  # fraction from 0 to 1

    print(position)
    # Update the moving elements based on the current position
    # apply the "easing" function to the requested position to adjust motion
    self._draw_position(easing(position))  # update the switch position

# update the switch value once the motion is complete
if (position >= 1) and not self._value:
    self._value = True
    break
if (
    position <= 0
) and self._value:  # ensures that the final position is drawn
    self._value = False
    break
```

The-Debarghya:
Now I get it, thanks a lot!

FoamyGuy (approved):
Looks good to me. I tested successfully on Feather ESP32-S2 TFT with this code:
```
import time
import board
import displayio
from adafruit_displayio_layout.widgets.switch_round import SwitchRound as Switch

display = board.DISPLAY

# Create the switch
my_switch = Switch(20, 30, animation_time=0)

my_group = displayio.Group()
my_group.append(my_switch)

# Add my_group to the display
display.show(my_group)

# Start the main loop
while True:
    time.sleep(1)
    my_switch.value = True
    time.sleep(1)
    my_switch.value = False
```

PR #77, opened 2022-11-05
Reintroduce graphviz

tekktrik:
Needed to build `sphinx` documentation, and was in the original CI files for this library.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,5 +10,8 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
+    - name: Load graphviz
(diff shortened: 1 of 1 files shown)


jepler (approved):
if we have many exceptions like this we'll want to do something less "one-off" but OK for now!

tekktrik:
This seemed to be the only one that I could find, and for what it's worth, we could resolve it be doing a mock import for the sphinx documentation I think if we wanted.

jepler:
I think it's used during the doc build to generate an inheritance diagram .. `sphinx.ext.inheritance_diagram`

PR #78, opened 2022-12-09
shorten very long file names

dhalbert:
The hotplug examples had very long filenames that caused problems on Windows for circup, possibly because the total pathlength was >260 or so.


Neradoc (approved):
LGTM

PR #79, opened 2023-01-31
adding version information to display in files packages

jposada202020:
Adding, the version information to the files, this way when building the packages information will be show in the py files.


tekktrik (approved):
Thanks!

PR #83, opened 2023-02-06
Removing Vectorshape Usage

jposada202020:
After CircuitPython 7.0.0 the usage of vectorshape has been deprecated. The current library has a rectangle_helper that uses it. 
https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/blob/29fe3f0bdda917c6c8efbdcba94885eeb6963fa3/adafruit_displayio_layout/widgets/__init__.py#L62
However after CP 7.0 this function becomes obsolete as the API for vectorio needs to add the shape to the group using the corresponding palette. Open for discussion, that having only the fill_area from bitmaptools will be easier than refactoring the cartesian to draw with both libraries. See documentation for vectorio.recatngle. So maybe there is a workaround but I could not find it during my tests. 

```py
group = displayio.Group()

palette = displayio.Palette(1)
palette[0] = 0x125690

rectangle = vectorio.Rectangle(pixel_shader=palette, width=40, height=30, x=55, y=45)
group.append(rectangle)
```

Cartesian is the only library using this, and is already hardcoded to use bitmaptool library only. 
https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/blob/29fe3f0bdda917c6c8efbdcba94885eeb6963fa3/adafruit_displayio_layout/widgets/cartesian.py#L360

### DIFF
--- a/adafruit_displayio_layout/widgets/__init__.py
+++ b/adafruit_displayio_layout/widgets/__init__.py
@@ -58,10 +58,10 @@ def rectangle_helper(
     if bitmaptool:
         bitmaptools.fill_region(bitmap, x0, y0, x0 + width, y0 + height, color_index)
     else:
-        rect = vectorio.Rectangle(width, height)
-        vectorio.VectorShape(
-            shape=rect,
+        vectorio.Rectangle(
             pixel_shader=palette,
+            width=width,
+            height=height,
             x=x0,
             y=y0,
         )


FoamyGuy:
Thank you for the fix @jposada202020. 

I tested this out on a Feather ESP32-S2 TFT. The changed version does no longer crash. But I noticed that it doesn't actually draw the rectangle either. 

I think it's because it creates a vectorio rectangle object but doesn't return it or do anything else with it, so it's never added to the display or shown. 

I think I am in favor of removing the `bitmaptool` argument altogether instead of fixing it to work with vectorio optionally. The only current usage of this helper has it hardcoded for bitmaptools anyhow, and the function would end up being kind of awkward if it had to return or add a vectorio shape since it behaves differently from the bitmap behavior which can "draw into" a given bitmap without needing to return or do anything else with it.

jposada202020:
Yes, Agreed with you, I have even tried to create a new palette object and the color_index passed to the function with no luck. I will work on the PR  for the cartesian widget this week and put this one as a draft. Thanks :)

jposada202020:
see #84

PR #84, opened 2023-02-06
Updating cartesian widget

jposada202020:
This will remove the usage of vectorshape usage and the rectangle_helper function in the `__init_.py` file.

tested with the three cartesian examples ,however a second pair of eyes would be helpful! :) tks


tekktrik:
Tagged you @FoamyGuy since you looked at the predecessor PR, but I can review too if you need.

FoamyGuy (approved):
Looks good to me. Tested successfully on PyPortal Titano.

Thank you @jposada202020!

PR #85, opened 2023-05-03
Missing type annotations, refs #50

### LINKED ISSUE #50: Missing Type Annotations
There are missing type annotations for some functions in this library.

The `typing` module does not exist on CircuitPython devices so the import needs to be wrapped in try/except to catch the error for missing import. There is an example of how that is done here:

```python
try:
    from typing import List, Tuple
except ImportError:
    pass
```

Once imported the typing annotations for the argument type(s), and return type(s) can be added to the function signature. Here is an example of a function that has had this done already:

```python
def wrap_text_to_pixels(
    string: str, max_width: int, font=None, indent0: str = "", indent1: str = ""
) -> List[str]:
```
If you are new to Git or Github we have a guide about contributing to our projects here: https://learn.adafruit.com/contribute-to-circuitpython-with-git-and-github

There is also a guide that covers our CI utilities and how to run them locally to ensure they will pass in Github Actions here: https://learn.adafruit.com/creating-and-sharing-a-circuitpython-library/check-your-code In particular the pages: `Sharing docs on ReadTheDocs` and `Check your code with pre-commit` contain the tools to install and commands to run locally to run the checks.

If you are attempting to resolve this issue and need help, you can post a comment on this issue and tag both @foamyguy and @kattni or reach out to us on Discord: https://adafru.it/discord in the `#circuitpython-dev` channel.

The following locations are reported by mypy to be missing type annotations:

- [ ] adafruit_displayio_layout/widgets/control.py:57
- [ ] adafruit_displayio_layout/widgets/control.py:85
- [ ] adafruit_displayio_layout/widgets/easing.py:79
- [ ] adafruit_displayio_layout/widgets/easing.py:87
- [ ] adafruit_displayio_layout/widgets/easing.py:95
- [ ] adafruit_displayio_layout/widgets/easing.py:105
- [ ] adafruit_displayio_layout/widgets/easing.py:115
- [ ] adafruit_displayio_layout/widgets/easing.py:123
- [ ] adafruit_displayio_layout/widgets/easing.py:134
- [ ] adafruit_displayio_layout/widgets/easing.py:145
- [ ] adafruit_displayio_layout/widgets/easing.py:153
- [ ] adafruit_displayio_layout/widgets/easing.py:164
- [ ] adafruit_displayio_layout/widgets/easing.py:175
- [ ] adafruit_displayio_layout/widgets/easing.py:183
- [ ] adafruit_displayio_layout/widgets/easing.py:194
- [ ] adafruit_displayio_layout/widgets/easing.py:205
- [ ] adafruit_displayio_layout/widgets/easing.py:213
- [ ] adafruit_displayio_layout/widgets/easing.py:221
- [ ] adafruit_displayio_layout/widgets/easing.py:229
- [ ] adafruit_displayio_layout/widgets/easing.py:237
- [ ] adafruit_displayio_layout/widgets/easing.py:247
- [ ] adafruit_displayio_layout/widgets/easing.py:257
- [ ] adafruit_displayio_layout/widgets/easing.py:267
- [ ] adafruit_displayio_layout/widgets/easing.py:279
- [ ] adafruit_displayio_layout/widgets/easing.py:291
- [ ] adafruit_displayio_layout/widgets/easing.py:299
- [ ] adafruit_displayio_layout/widgets/easing.py:309
- [ ] adafruit_displayio_layout/widgets/easing.py:322
- [ ] adafruit_displayio_layout/widgets/easing.py:330
- [ ] adafruit_displayio_layout/widgets/easing.py:341
- [ ] adafruit_displayio_layout/widgets/easing.py:352
- [ ] adafruit_displayio_layout/widgets/easing.py:359
- [ ] adafruit_displayio_layout/widgets/easing.py:372
- [ ] adafruit_displayio_layout/widgets/widget.py:167
- [ ] adafruit_displayio_layout/widgets/widget.py:198
- [ ] adafruit_displayio_layout/widgets/widget.py:278
- [ ] adafruit_displayio_layout/widgets/widget.py:295
- [ ] adafruit_displayio_layout/layouts/grid_layout.py:51
- [ ] adafruit_displayio_layout/layouts/grid_layout.py:290
- [ ] adafruit_displayio_layout/layouts/grid_layout.py:308
- [ ] adafruit_displayio_layout/widgets/switch_round.py:420
- [ ] adafruit_displayio_layout/widgets/switch_round.py:664
- [ ] adafruit_displayio_layout/widgets/switch_round.py:686
- [ ] adafruit_displayio_layout/widgets/switch_round.py:783
- [ ] adafruit_displayio_layout/widgets/switch_round.py:803
- [ ] adafruit_displayio_layout/widgets/switch_round.py:827
- [ ] adafruit_displayio_layout/widgets/switch_round.py:837
- [ ] adafruit_displayio_layout/widgets/switch_round.py:849
- [ ] adafruit_displayio_layout/widgets/switch_round.py:854
- [ ] adafruit_displayio_layout/widgets/switch_round.py:887
- [ ] adafruit_displayio_layout/widgets/switch_round.py:904
- [ ] adafruit_displayio_layout/widgets/icon_widget.py:54
- [ ] adafruit_displayio_layout/widgets/icon_widget.py:87
- [ ] adafruit_displayio_layout/widgets/flip_input.py:84
- [ ] adafruit_displayio_layout/widgets/flip_input.py:329
- [ ] adafruit_displayio_layout/widgets/flip_input.py:431
- [ ] adafruit_displayio_layout/widgets/flip_input.py:447
- [ ] adafruit_displayio_layout/widgets/flip_input.py:506
- [ ] adafruit_displayio_layout/widgets/flip_input.py:525
- [ ] adafruit_displayio_layout/widgets/flip_input.py:580
- [ ] adafruit_displayio_layout/widgets/flip_input.py:622
- [ ] adafruit_displayio_layout/widgets/icon_animated.py:77
- [ ] adafruit_displayio_layout/widgets/icon_animated.py:131
- [ ] adafruit_displayio_layout/widgets/icon_animated.py:168
- [ ] adafruit_displayio_layout/widgets/icon_animated.py:263

cguardia:
This PR adds type declarations and mypy compliance to this package, as discussed on issue #50. The files were tested with mypy running in strict mode, to make sure no type declaration issues remained. In some cases, it was necessary to modify the code to insure compliance.

Sorry for the long commit. I grabbed the first free issue at the sprint and turned out to be a larger undertaking than expected.

### DIFF
--- a/adafruit_displayio_layout/layouts/grid_layout.py
+++ b/adafruit_displayio_layout/layouts/grid_layout.py
@@ -24,7 +24,7 @@ Implementation Notes
 """
 try:
     # Used only for typing
-    from typing import Any, List, Tuple, Union
+    from typing import Any, List, Optional, Tuple, Union
 except ImportError:
     pass
 
@@ -381,7 +381,7 @@ class GridLayout(displayio.Group):
         cell_content: displayio.Group,
         grid_position: Tuple[int, int],
         cell_size: Tuple[int, int],
-        cell_anchor_point: Union[Tuple[float, ...], None] = None,
+        cell_anchor_point: Optional[Tuple[float, ...]] = None,
     ) -> None:
         """Add a child to the grid.
 
--- a/adafruit_displayio_layout/layouts/grid_layout.py
+++ b/adafruit_displayio_layout/layouts/grid_layout.py
@@ -24,7 +24,7 @@ Implementation Notes
 """
 try:
     # Used only for typing
-    from typing import Tuple
+    from typing import Any, List, Tuple, Union
 except ImportError:
     pass
 
@@ -61,18 +61,18 @@ class GridLayout(displayio.Group):
     # pylint: disable=too-many-instance-attributes
     def __init__(
         self,
-        x,
-        y,
-        width,
-        height,
-        grid_size,
-        cell_padding=0,
-        divider_lines=False,
-        h_divider_line_rows=None,
-        v_divider_line_cols=None,
-        divider_line_color=0xFFFFFF,
-        cell_anchor_point=(0.0, 0.0),
-    ):
+        x: int,
+        y: int,
+        width: int,
+        height: int,
+        grid_size: tuple[int, int],
+        cell_padding: int = 0,
+        divider_lines: bool = False,
+        h_divider_line_rows: Union[Tuple[int, ...], List[int], None] = None,
+        v_divider_line_cols: Union[Tuple[int, ...], List[int], None] = None,
+        divider_line_color: int = 0xFFFFFF,
+        cell_anchor_point: Tuple[float, float] = (0.0, 0.0),
+    ) -> None:
         super().__init__(x=x, y=y)
         self.x = x
         self.y = y
@@ -80,13 +80,13 @@ class GridLayout(displayio.Group):
         self._height = height
         self.grid_size = grid_size
         self.cell_padding = cell_padding
-        self._cell_content_list = []
+        self._cell_content_list: List[dict[str, Any]] = []
         self._cell_anchor_point = cell_anchor_point
 
-        self._divider_lines = []
+        self._divider_lines: List[dict[str, Any]] = []
         self._divider_color = divider_line_color
-        self.h_divider_line_rows = h_divider_line_rows
-        self.v_divider_line_cols = v_divider_line_cols
+        self.h_divider_line_rows = h_divider_line_rows or tuple()
+        self.v_divider_line_cols = v_divider_line_cols or tuple()
 
         self._divider_lines_enabled = (
             (divider_lines is True)
@@ -95,19 +95,14 @@ class GridLayout(displayio.Group):
         )
 
         if divider_lines:
-            if self.h_divider_line_rows is None:
+            if h_divider_line_rows is None:
                 self.h_divider_line_rows = []
                 for _y in range(self.grid_size[1] + 1):
                     self.h_divider_line_rows.append(_y)
-            if self.v_divider_line_cols is None:
+            if v_divider_line_cols is None:
                 self.v_divider_line_cols = []
                 for _x in range(self.grid_size[0] + 1):
                     self.v_divider_line_cols.append(_x)
-        else:
-            if not h_divider_line_rows:
-                self.h_divider_line_rows = tuple()
-            if not v_divider_line_cols:
-                self.v_divider_line_cols = tuple()
 
         # use at least 1 padding so that content is inside the divider lines
         if cell_padding == 0 and (
@@ -115,7 +110,7 @@ class GridLayout(displayio.Group):
         ):
             self.cell_padding = 1
 
-    def _layout_cells(self):
+    def _layout_cells(self) -> None:
         # pylint: disable=too-many-locals, too-many-branches, too-many-statements
         for cell in self._cell_content_list:
             if cell["content"] not in self:
@@ -382,8 +377,12 @@ class GridLayout(displayio.Group):
                         self.append(line_obj["tilegrid"])
 
     def add_content(
-        self, cell_content, grid_position, cell_size, cell_anchor_point=None
-    ):
+        self,
+        cell_content: displayio.Group,
+        grid_position: Tuple[int, int],
+        cell_size: Tuple[int, int],
+        cell_anchor_point: Union[Tuple[float, ...], None] = None,
+    ) -> None:
         """Add a child to the grid.
 
         :param cell_content: the content to add to this cell e.g. label, button, etc...
@@ -412,7 +411,7 @@ class GridLayout(displayio.Group):
         self._cell_content_list.append(sub_view_obj)
         self._layout_cells()
 
-    def get_cell(self, cell_coordinates):
+    def get_cell(self, cell_coordinates: Tuple[int, int]) -> displayio.Group:
         """
         Return a cells content based on the cell_coordinates. Raises
         KeyError if coordinates were not found in the GridLayout.
--- a/adafruit_displayio_layout/widgets/cartesian.py
+++ b/adafruit_displayio_layout/widgets/cartesian.py
@@ -36,7 +36,7 @@ except ImportError:
     pass
 
 try:
-    from typing import Tuple
+    from typing import Any, List, Optional, Tuple
 except ImportError:
     pass
 
@@ -178,7 +178,7 @@ class Cartesian(Widget):
         tick_color: int = 0xFFFFFF,
         major_tick_stroke: int = 1,
         major_tick_length: int = 5,
-        tick_label_font=terminalio.FONT,
+        tick_label_font: terminalio.FONT = terminalio.FONT,
         font_color: int = 0xFFFFFF,
         pointer_radius: int = 1,
         pointer_color: int = 0xFFFFFF,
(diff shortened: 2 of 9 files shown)


adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -61,32 +61,32 @@ class GridLayout(displayio.Group):
...
-    ):
+        x: int,
+        y: int,
+        width: int,
+        height: int,
+        grid_size: tuple[int, int],
+        cell_padding: int = 0,
+        divider_lines: bool = False,
+        h_divider_line_rows: Union[Tuple[int, ...], List[int], None] = None,
+        v_divider_line_cols: Union[Tuple[int, ...], List[int], None] = None,
```
FoamyGuy, line 72: For arguments that allow `None` we've been using `Optional` instead of `Union` with None as one of the choices. However since this one already is a Union and to me they convey the same meaning I think it's fine to keep it this way here. @tekktrik you have any thoughts on this?

adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -95,27 +95,22 @@ def __init__(
...
-            if self.v_divider_line_cols is None:
+            if v_divider_line_cols is None:
                 self.v_divider_line_cols = []
                 for _x in range(self.grid_size[0] + 1):
                     self.v_divider_line_cols.append(_x)
-        else:
-            if not h_divider_line_rows:
-                self.h_divider_line_rows = tuple()
-            if not v_divider_line_cols:
-                self.v_divider_line_cols = tuple()
```
FoamyGuy, line 110: The way the divider line handling has been refactored looks functionally the same to me, but I do think it's worth doublechecking the behavior on a device. I'll test out everything I can and report back if anything comes up.

adafruit_displayio_layout/layouts/grid_layout.py
```diff
@@ -382,8 +377,12 @@ def _layout_cells(self):
                         self.append(line_obj["tilegrid"])
 
     def add_content(
-        self, cell_content, grid_position, cell_size, cell_anchor_point=None
-    ):
+        self,
+        cell_content: displayio.Group,
+        grid_position: Tuple[int, int],
+        cell_size: Tuple[int, int],
+        cell_anchor_point: Union[Tuple[float, ...], None] = None,
```
FoamyGuy: Here for `cell_anchor_point` I do think `Optional` may be a better choice for the type to convey that it can be None, since there is only 1 other valid type we can also get rid of Union.
```
cell_anchor_point: Optional[Tuple[float, ...]] = None,
```

adafruit_displayio_layout/widgets/cartesian.py
```diff
@@ -609,6 +616,9 @@ def add_plot_line(self, x: int, y: int) -> None:
 
         rtype: None
         """
+        # mypy, plot_line_point can't be None at this point
+        assert self.plot_line_point is not None
```
FoamyGuy: I think it's worth wider consideration and conversation about including the assert statements. If I'm understanding correctly, this code resolves an issue that mypy reports.

We haven't enabled automated mypy yet, and when we do we'll decide on a configuration used with regards to which warnings will constitute pass or fail of the actions run. 

It's worth considering whether we want to include asserts like this in order to quiet certain errors and warnings. In my understanding I think this line will add to the size of the mpy file and RAM used when the library is imported, both of which are fairly constrained on some devices. In contrast the typing annotations get stripped out as far as I understand, so those do not end up occupying any additional space once compiled to mpy.

adafruit_displayio_layout/widgets/flip_input.py
```diff
@@ -508,7 +523,9 @@ def value(self):
         return self._value
 
     @value.setter
-    def value(self, new_value):  # Set the value based on the index or on the string.
+    def value(
+        self, new_value: int | str
```
FoamyGuy, line 527: I don't know if I've ever seen the pipe syntax before. Is that different than `Union` functionally? or just a different way to write the same thing?

adafruit_displayio_layout/widgets/icon_animated.py
```diff
@@ -25,12 +25,17 @@
...
 import bitmaptools
-from displayio import TileGrid, Bitmap, Palette
+from displayio import TileGrid, Bitmap, Display, Palette
 import adafruit_imageload
 from adafruit_displayio_layout.widgets.icon_widget import IconWidget
 from adafruit_displayio_layout.widgets.easing import quadratic_easeout as easein
 from adafruit_displayio_layout.widgets.easing import quadratic_easein as easeout
 
+try:
+    from typing import Any, Optional, Tuple
```
FoamyGuy, line 35: I think we want the display import to be in this try block after the typing import. 
```
from displayio import Display
```

This way the import will not occur on microcontrollers which don't have `typing` so we will save a bit of RAM by not importing the class that won't be used at run time.

adafruit_displayio_layout/widgets/widget.py
```diff
@@ -166,14 +172,14 @@ class Widget(displayio.Group):
...
-        y=0,
-        scale=1,
-        width=None,
-        height=None,
-        anchor_point=None,
-        anchored_position=None,
-    ):
+        x: int = 0,
+        y: int = 0,
+        scale: float = 1,
```
FoamyGuy: this `scale` gets passed to displayio.Group constructor. I believe it can only use `int` not `float` for valid values.

FoamyGuy (changes requested):
Thanks for working on this @cguardia.

I have a few suggested changes and some questions that I think are worth considering with the wider team. I'd also like to give this a try on a device since there are some changes that go beyond just the type annotations albeit relatively minor, still worth double checking the behavior.

adafruit_displayio_layout/layouts/grid_layout.py
tekktrik: Optional is equivalent to Union with None, so no problem there.  I think we've traditionally used Optional, but I actually think once the pipe operator (|) was implemented to replace Union, the standard has actually become `[type of thing] | None`.  I would argue as long as None is allowed by the annotation it should be good.

adafruit_displayio_layout/widgets/flip_input.py
tekktrik: @foamyguy I mentioned it in my previous response but it's syntax that equates to using Union.  It is not available in Python 3.8 however.

FoamyGuy (commented):
After discussion during the meeting I tested out some alternative approaches that allow us to resolve the mypy errors without needing to have the assert lines. I'll push those tweaks to this branch

adafruit_displayio_layout/widgets/cartesian.py
cguardia: I'm really not sure how code is run on devices, but ``assert`` statements are ignored when using the -o flag or setting ``__debug__`` environment variable to ``False``.

cguardia:
@FoamyGuy I'm OK with using the alternative approach to remove assert statements. I made some requested changes, but not these, because you indicated you are already on it. Let me know what else I can do to advance this PR. Thanks,

FoamyGuy (approved):
The latest version looks good to me. Thanks again for taking this on @cguardia! 

I pushed one commit to tweak some of the Display asserts to check for not None instead of Display in order to avoid needing the Display import at runtime. I confirmed that change is acceptable to mypy compared to the Display isinstance check.

I tested all of the micro-controller examples successfully on a PyPortal Titano.

PR #88, opened 2023-11-01
remove displayio.Shape usage and replace with vectorio.Rectangle

FoamyGuy:
This change is necessary for CircuitPython 9.0.0+ because `displayio.Shape` was removed from the core.

I tested the updated version successfully with a PyPortal Titano on `9.0.0-alpha.2`


tannewt (approved):
Thank you!

PR #92, opened 2023-12-16
Gridlayout cell contains

FoamyGuy:
This PR adds a few main things:

- new properties for `width` and `height` that turned out to be convenient for the real functionality I wanted to add
- enhancement to `get_cell()` to allow it to correctly locate multi-spanned cells
- new function `which_cell_contains()` which takes a pixel x,y location and returns the cell coordinates of the cell that contains that pixel location. Helpful for use with touchscreens.
- added terminalio to doc mock list to get docs building successfully

### DIFF
--- a/adafruit_displayio_layout/layouts/grid_layout.py
+++ b/adafruit_displayio_layout/layouts/grid_layout.py
@@ -448,7 +448,7 @@ class GridLayout(displayio.Group):
     @property
     def height(self) -> int:
         """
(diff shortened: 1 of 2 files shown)


FoamyGuy:
The latest commit to this branch fixes an issue that resulted from the divider line objects being erroneously removed from their containing group twice which raised an exception.

tannewt (approved):
Looks good to me!

PR #95, opened 2025-01-05
Remove '__version__' and '__repo__' to save RAM

SamantazFox:
As said in the title, this removes the `__version__` and `__repo__` symbols from all modules to save up on RAM.

During my testing on an Feather RP2040 CAN, I was able to save 2.5 kB of that precious RAM in a worst-case scenario (= all modules being imported). This represents almost 1% of that MCU's total amount.

Notes: to get the numbers below, I applied the same optimization to [Adafruit_CircuitPython_imageload](https://github.com/adafruit/Adafruit_CircuitPython_imageload), on which this module depends, and all the libraries were compiled to `.mpy` beforehand using `mpy-cross` v9.2.1, which is the exact CP version that my board is running.

Code used to test:
```python
import gc

from adafruit_displayio_layout import *
from adafruit_displayio_layout.layouts.grid_layout import *
from adafruit_displayio_layout.layouts.linear_layout import *
from adafruit_displayio_layout.layouts.page_layout import *
from adafruit_displayio_layout.layouts.tab_layout import *
from adafruit_displayio_layout.widgets.cartesian import *
from adafruit_displayio_layout.widgets.control import *
from adafruit_displayio_layout.widgets.easing import *
from adafruit_displayio_layout.widgets.flip_input import *
from adafruit_displayio_layout.widgets.icon_animated import *
from adafruit_displayio_layout.widgets.icon_widget import *
from adafruit_displayio_layout.widgets.switch_round import *
from adafruit_displayio_layout.widgets.widget import *

print(gc.mem_alloc())
```

results:
* before: 59792 bytes
* after:  57216 bytes
* saved: 2576 bytes

### DIFF
--- a/adafruit_displayio_layout/layouts/grid_layout.py
+++ b/adafruit_displayio_layout/layouts/grid_layout.py
@@ -32,9 +32,6 @@ import math
 import displayio
 from vectorio import Rectangle
 
-__version__ = "0.0.0+auto.0"
-__repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout.git"
-
 
 class GridLayout(displayio.Group):
     """
--- a/adafruit_displayio_layout/layouts/page_layout.py
+++ b/adafruit_displayio_layout/layouts/page_layout.py
@@ -32,9 +32,6 @@ except ImportError:
 
 import displayio
 
-__version__ = "0.0.0+auto.0"
-__repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout.git"
-
 
 class PageLayout(displayio.Group):
     """
--- a/adafruit_displayio_layout/layouts/tab_layout.py
+++ b/adafruit_displayio_layout/layouts/tab_layout.py
@@ -37,9 +37,6 @@ from adafruit_display_text.bitmap_label import Label
 from adafruit_imageload.tilegrid_inflator import inflate_tilegrid
 from adafruit_displayio_layout.layouts.page_layout import PageLayout
 
-__version__ = "0.0.0+auto.0"
-__repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout.git"
-
 
 class TabLayout(displayio.Group):
     """
--- a/adafruit_displayio_layout/widgets/cartesian.py
+++ b/adafruit_displayio_layout/widgets/cartesian.py
@@ -41,10 +41,6 @@ except ImportError:
     pass
 
 
-__version__ = "0.0.0+auto.0"
-__repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout.git"
-
-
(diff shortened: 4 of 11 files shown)


tannewt:
Does this work with circup still? I think circup uses this info to map back to a package but I could be wrong.

Neradoc:
We do need the version somewhere so it can be read by code, but mostly tools like circup for updates.
I guess since it's common to all modules in the package, in could be in a single file like the `__init__.py` at the root level of the package. Maybe it could even be in something like a "version.py" file that doesn't get loaded automatically when using the library (`__init__.py` is) though a version string is not that big.
I don't know if there is a way to have it in some metadata in the file that wouldn't be loaded to memory.

The repository URL is read in circup, if present, but used only in one place, when warning that a module has a bad version number, adding `Please file an issue in the library repo.` Circup could use the bundle json file which does contain that information too.

Also, if we do any of that, it's worth looking at making it a policy across the whole bundle.

SamantazFox:
> We do need the version somewhere so it can be read by code, but mostly tools like circup for updates.

Not sure if that's relevant, but from what I've seen in the bundle, the presence of these two symbols is quite inconsistent. For instance it's absent from [`layouts/linear_layout.py`](https://github.com/adafruit/Adafruit_CircuitPython_DisplayIO_Layout/blob/5b80e0d27e7fb4f7921a6f1587d65a799024ce6b/adafruit_displayio_layout/layouts/linear_layout.py).

> I guess since it's common to all modules in the package, in could be in a single file like the `__init__.py` at the root level of the package. Maybe it could even be in something like a "version.py" file that doesn't get loaded automatically when using the library (`__init__.py` is) though a version string is not that big. I don't know if there is a way to have it in some metadata in the file that wouldn't be loaded to memory.

In my opinion, some root-level `version.[json,txt]` would be great, especially since it's easy to distinguish from actual code (so it can be easily removed to save on flash storge) however that wouldn't work for single file packages (e.g: `adafruit_ticks`).

What do you think of embedding these in the comments, beneath the SPDX license lines?

> Also, if we do any of that, it's worth looking at making it a policy across the whole bundle.

I was wondering, is there a place to discuss issues that affect the whole bundle or multiple packages at once?

Neradoc:
> Not sure if that's relevant, but from what I've seen in the bundle, the presence of these two symbols is quite inconsistent.

Yeah I agree.

> What do you think of embedding these in the comments, beneath the SPDX license lines?

Comments are deleted from mpy files. And actually I think other libraries might want to check the version. So I don't think we can avoid having the version in memory once. But getting rid of the URL and duplicate version strings would be nice for M0 boards when that matters.

> I was wondering, is there a place to discuss issues that affect the whole bundle or multiple packages at once?

The standard way would be to open an issue in the Circuitpython library with the `library` tag.
https://github.com/adafruit/circuitpython/issues?q=is%3Aopen+is%3Aissue+label%3Alibrary

FoamyGuy:
This could be a good topic to discus in the weeds during a meeting as well.

PR #96, opened 2025-02-12
 Update how board.DISPLAY is checked

jepler:
Historically, boards with no built-in display have not had a `board.DISPLAY` property.

Soon, with https://github.com/adafruit/circuitpython/pull/10028, any board with displayio will have a `board.DISPLAY` property. This property will be `None` if no display is configured, or a truthy non-`None` value if a display is configured (including dynamically at runtime).

This revised check will work on both old and new circuitpython versions.


jepler:
Over in circuitpython we ended up going with `supervisor.runtime.display` instead of `board.DISPLAY` for the new functionality. As such, this PR is not useful to merge after all.

PR #99, opened 2025-04-02
add pop_content() rename get_cell() to get_content()

FoamyGuy:
This adds a new `pop_content()` function to GridLayout. Previously there was no easy way to remove content from the grid once it was added. This functionality made the Set card game I am working on much easier to implement. 

I've also rename `get_cell()` to `get_content()` which is a breaking change, we'll need to bump major version number if merged. In hindsight I think the add and get functions should have matched in the terms used in their names. "content" makes more sense than "cell" to me since what it gets or adds is the content that into the cell at the position specified.


tannewt (approved):
Thank you!

Issue #104, opened 2025-07-13
Example File Neede Updating: get_cell() example needs to be changed to get_content()


gh3262:
Example file: displayio_layout_grid_layout_get_cell_test.py uses 'get_cell()' however this was changed in #99 to 'get_content()' so the example needs to be updated.

PR #105, opened 2025-09-02
Fix example to use get_content() instead of get_cell()

shubham0x13:
Replaced `get_cell()` with `get_content()` in `displayio_layout_grid_layout_get_cell_test.py` to align with the API changes introduced in #99. This resolves issue #104.


FoamyGuy (approved):
Looks good, thanks for the fix!

